PR #7, opened 2016-11-10
Renames census to instrumentation.stats in package names

dinooliva:


census.proto
```diff
@@ -13,7 +13,7 @@ syntax = "proto3";
 
 package google.census;
```
sebright: Should the packages be updated, too?

dinooliva: This is overridden in Java by the java_packagename spec.

PR #17, opened 2017-02-07
Changes tags in StatsContext proto from string to bytes.

dinooliva:
And documents encoding format.

### DIFF
--- a/stats/stats_context.proto
+++ b/stats/stats_context.proto
@@ -16,6 +16,17 @@ package google.instrumentation;
 option java_package = "com.google.instrumentation.stats.proto";
 option java_outer_classname = "StatsContextProto";
 
+// Wire format of instrumentation stats (aka Census) tags.
 message StatsContext {
-  string tags = 1;
+  // Encodes all stats tags (aka Census tags) to propagate over an RPC in a
+  // single byte vector.
+  //
+  // The format is:
+  //  num_tags [key_len key_bytes value_len value_bytes]*
+  //
+  // All implementations must honor these specs:
+  //  * The num_tags is represented using varint.
+  //  * The key/value lengths are represented using varint.
+  //  * keys/values do not include the terminating character '\0'.
+  bytes tags = 1;
 }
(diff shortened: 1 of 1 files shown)


stats/stats_context.proto
```diff
@@ -16,6 +16,17 @@ package google.instrumentation;
...
 option java_outer_classname = "StatsContextProto";
 
+// Wire format of instrumentation stats (aka Census) tags.
 message StatsContext {
-  string tags = 1;
+  // Encodes all stats tags (aka Census tags) to propagate over an RPC in a
+  // single byte vector.
+  //
+  // The format is:
+  //  num_tags [key_len key_bytes value_len value_bytes]*
```
a-veitch: As we talked about IRL, I don't think we need the num_tags. But lets talk about this some more.

stats/stats_context.proto
```diff
@@ -16,6 +16,17 @@ package google.instrumentation;
...
+  //
+  // All implementations must honor these specs:
+  //  * The num_tags is represented using varint.
+  //  * The key/value lengths are represented using varint.
+  //  * keys/values do not include the terminating character '\0'.
```
a-veitch: I don't think we need this condition. In fact, I think we should state that tag keys/values must be in UTF-8 format (another possibility is US-ASCII, but I think we will have cases where people want UTF-8, especially for values).

stats/stats_context.proto
```diff
@@ -16,6 +16,17 @@ package google.instrumentation;
...
 message StatsContext {
-  string tags = 1;
+  // Encodes all stats tags (aka Census tags) to propagate over an RPC in a
+  // single byte vector.
+  //
+  // The format is:
+  //  num_tags [key_len key_bytes value_len value_bytes]*
+  //
+  // All implementations must honor these specs:
+  //  * The num_tags is represented using varint.
```
a-veitch: Should provide a pointer to varint documentation, e.g. https://developers.google.com/protocol-buffers/docs/encoding#varints

dinooliva: Done.

stats/stats_context.proto
dinooliva: We last agreed upon US-ASCII, so I've added that to the requirements. We should review this in our next inter-language sync.

stats/stats_context.proto
dinooliva: Ack.

dinooliva: If we want num_tags, maybe it should be a different field in the proto.

PR #18, opened 2017-02-09
Release the span_context.proto definition.

bogdandrutu:


trace/trace_context.proto
```diff
@@ -0,0 +1,39 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package google.instrumentation.trace.proto;
+
+option java_multiple_files = true;
```
sebright: I don't know if this matters, but `stats_context.proto` uses nested classes instead of multiple files: https://github.com/bogdandrutu/instrumentation-proto/blob/575cdb257cf149a186dc2276dcdb4c3ffba9ff1a/stats/stats_context.proto#L17

bogdandrutu: I changed to use nested classes and changed the name to span_context.

sebright:
@bogdandrutu Has this proto format been replaced?

PR #22, opened 2017-03-01
Update stats_context.proto to match the new encodings.

songy23:


stats/stats_context.proto
```diff
@@ -16,20 +16,26 @@ package google.instrumentation;
...
-  //
-  // All implementations must honor these specs:
-  //  * The num_tags is represented using varint.
-  //  * The key/value lengths are represented using varint.
-  //  * All key/values must be US-ASCII.
-  //
-  // For details about varint, see:
-  //  https://developers.google.com/protocol-buffers/docs/encoding#varints
+  // Where:
+  //  * tag_type is one byte, and is used to describe the format of value_bytes.
```
dinooliva: I think this description needs to be changed slightly - I've added a suggestion in the shared document.

PR #23, opened 2017-03-13
add BUILD file for grpc

makdharma:
Add bazel rule to generate census proto library that can be used by
other repos that use gRPC.

### DIFF
--- a/stats/BUILD
+++ b/stats/BUILD
@@ -1,0 +1,25 @@
+# Copyright 2017, Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


bogdandrutu:
Do you need a BUILD file only for this? Why do you not use the main BUILD file from the root directory of this project.

makdharma:
The reason for not using main BUILD file has to do with bazel and grpc limitation (I've asked gRPC folks to file a bug). gRPC proto library will only build when the proto and BUILD files are in the same folder. I can move this BUILD file to the main BUILD when this limitation is removed.

sebright:
@makdharma It looks like the second BUILD file is causing a build failure.  Would it be possible to access this repository using [`new_http_archive`](https://bazel.build/versions/master/docs/be/workspace.html#new_http_archive) instead, and then specify the new BUILD file with the `build_file` field?

sebright:
@makdharma Is this change still needed for gRPC?

bogdandrutu:
Does this https://github.com/census-instrumentation/opencensus-proto/pull/38 resolve your issue?

sebright:
I'm closing this PR for now, since it isn't active and #34 added a build file in the stats directory.

PR #24, opened 2017-03-14
Add text-format proto for RPC constants.

songy23:
Translate Java-format [RPCConstants](https://github.com/google/instrumentation-java/blob/master/core/src/main/java/com/google/instrumentation/stats/RpcConstants.java) into text-format protobuf. The data structures and enum types are defined in [census.proto](https://github.com/google/instrumentation-proto/blob/master/stats/census.proto).


stats/rpc_constants.textpb
```diff
@@ -0,0 +1,770 @@
...
+    }
+    tag_keys: "OpStatus"
+    tag_keys: "method"
+  }
+
+  view_descriptor {
+    name: "rpc client roundtrip_latency"
+    description: "Latency in msecs"
+    measurement_descriptor_name: "/rpc/client/roundtrip_latency"
+    distribution_aggregation: {
```
dinooliva: These bucket boundaries get copied a lot - is there any way to factor them out?

songy23: I was also wondering about that and have searched for some examples, but by so far I haven't found a way to create reusable variables in .textpb. Some other user has also commented about this issue under text-proto documentation, but it seems no one knows how to do it (assign values to a variable to reduce string or list duplication).

a-veitch: There isn't a way to do that, unfortunately (short of writing a template, and running it through a preprocessor/build step).

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,770 @@
+# Constants for collecting rpc stats.
```
a-veitch: Need a longer, more descriptive comment here. Should summarize the rest of the file.

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,770 @@
...
+rpc_constants {
+  # Census defined rpc client MeasurementDescriptors.
```
a-veitch: s/Census defined rpc/RPC/

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,770 @@
...
+  measurement_descriptor {
+    name: "/rpc/client/error_count"
+    description: "RPC Errors"
```
a-veitch: Descriptions should be longer, e.g. "Count of client RPC's that received an error (unsuccessful) response"

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,770 @@
...
+    unit: {
+      power10: 0
```
a-veitch: No need to specify 0 values for power10 (this is the default if not explictly set).

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,770 @@
...
+    description: "Response bytes"
+    unit: {
+      power10: 0
+      numerators: BYTES
+    }
+  }
+
+  measurement_descriptor {
+    name: "/rpc/client/roundtrip_latency"
+    description: "RPC roundtrip latency msec"
```
a-veitch: Don't use as much shorthand in descriptions (s/msec/milliseconds/)

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,770 @@
...
+  measurement_descriptor {
+    name: "/rpc/client/uncompressed_response_bytes"
+    description: "Uncompressed Response bytes"
+    unit: {
+      power10: 0
+      numerators: BYTES
+    }
+  }
+
+  #Census defined rpc server MeasurementDescriptors.
```
a-veitch: Nit: need a space after #.

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,770 @@
...
+  }
+
+  # Census defined rpc client ViewDescriptors.
+  view_descriptor {
+    name: "rpc client error_count"
+    description: "RPC Errors"
+    measurement_descriptor_name: "/rpc/client/error_count"
+    distribution_aggregation: {
+    }
+    tag_keys: "OpStatus"
```
a-veitch: We should separately consider what tag key names we would like here. We have traditionally used OpStatus and method, but possibly being more explicit and consistent and using (e.g.) rpc_status and rpc_method would be better.

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,770 @@
...
+    tag_keys: "method"
+  }
+
+  view_descriptor {
+    name: "rpc client roundtrip_latency"
+    description: "Latency in msecs"
+    measurement_descriptor_name: "/rpc/client/roundtrip_latency"
+    distribution_aggregation: {
+      bucket_bounds: 0.0
```
a-veitch: You can make this a lot shorter:

bucket_bounds: [1.0, 2.0, 3.0, ... 100000.0]

stats/rpc_constants.textpb
a-veitch: Can also use list syntax for tag keys, and add comments, maybe something like:

tag_keys: [
  'OpStatus',  # Tag value will reflect status of operation
  'method' # Tag value will be RPC service/method name.
]

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,770 @@
...
+      bucket_bounds: 1.0
```
a-veitch: Should there be a lower bound of 0?

stats/rpc_constants.textpb
songy23: Yep, fixed.

stats/rpc_constants.textpb
songy23: Removed

stats/rpc_constants.textpb
songy23: Fixed

stats/rpc_constants.textpb
songy23: Fixed

stats/rpc_constants.textpb
songy23: Will add comments and mark as follow-up here

stats/rpc_constants.textpb
songy23: Great idea, updated

stats/rpc_constants.textpb
songy23: I think so, since the bucket boundary begins from [-infinity](https://github.com/google/instrumentation-proto/blob/master/stats/census.proto#L200)

stats/rpc_constants.textpb
songy23: Yep, updated with a more detailed summary

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,620 @@
...
+}
+
+measurement_descriptor {
+  name: "/rpc/server/server_latency"
+  description: "Latency in milliseconds on server RPC side"
+  unit: {
+    power10: -3
+    numerators: SECONDS
+  }
+}
```
zhangkun83: What's the difference between the above two?
I asked the same question internally, but didn't get a good answer.

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,620 @@
...
+  name: "/rpc/server/finished_count"
+  description: "Number of server RPCs finished"
+  unit: {
+    numerators: SCALAR
+  }
+}
+
+# Census defined RPC client ViewDescriptors.
+view_descriptor {
+  name: "rpc client error_count"
```
zhangkun83: I know this is inherited from RpcConstants, but the spaces still look out of place. Is it the convention to use spaces?

zhangkun83: Probably better to reflect the type of the view (distribution vs. interval) in the name.

songy23: Yep, we are discussing on naming convention on team's doc. I'll update the names once we came to a conclusion.

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,620 @@
...
+    numerators: BYTES
+  }
+}
+
+# Census defined RPC server MeasurementDescriptors.
+measurement_descriptor {
+  name: "grpc.io/server/error_count"
+  description: "Count of server RPCs that received an error (unsuccessful) response"
+  unit: {
+    numerators: BYTES
```
sebright: `error_count` should be `SCALAR`.

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,620 @@
...
+
+measurement_descriptor {
+  name: "grpc.io/server/finished_count"
+  description: "Number of server RPCs finished"
+  unit: {
+    numerators: SCALAR
+  }
+}
+
+# Census defined RPC client ViewDescriptors.
```
sebright: I think this should be "client distribution ViewDescriptors".

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,620 @@
...
+      0.0, 1024.0, 2048.0, 4096.0, 16384.0, 65536.0, 262144.0, 1048576.0, 4194304.0,
+      16777216.0, 67108864.0, 268435456.0, 1073741824.0, 4294967296.0
+    ]
+  }
+  tag_keys: [
+    "method" # Tag value will be the name of server RPC method/service
+  ]
+}
+
+# Census defined RPC client IntervalViewDescriptors.
```
sebright: I think this should have a space, i.e., "client interval ViewDescriptors", because we don't have a separate `IntervalViewDescriptor` message.

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,620 @@
+# Constants for collecting specific RPC stats, including:
+#   - Measurement Descriptor, which describes a data point (measurement) type.
+#   - View Descriptor, which specifies an AggregationDescriptor and a set of tag keys.
+#     View Descriptor has one of the two kinds of Aggregation types associated with the View,
+#     Interval Aggregation Descriptor or Distribution Aggregation Descriptor.
+#     Aggregations are broken down by the unique set of matching tag values for each measurement.
+
+# Census defined RPC client MeasurementDescriptors.
```
sebright: Should this say "gRPC defined" instead of "Census defined", since all of the names contain "grpc.io"?

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,620 @@
...
+      0.0, 1024.0, 2048.0, 4096.0, 16384.0, 65536.0, 262144.0, 1048576.0, 4194304.0,
+      16777216.0, 67108864.0, 268435456.0, 1073741824.0, 4294967296.0
+    ]
+  }
+  tag_keys: [
+    "method" # Tag value will be the name of client RPC method/service
+  ]
+}
+
+# Census defined RPC server ViewDescriptors.
```
sebright: This should also say "distribution".

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,620 @@
...
+
+# Census defined RPC client IntervalViewDescriptors.
+view_descriptor {
+  name: "grpc.io/client/roundtrip_latency/interval"
+  description: "Minute and Hour stats for latency in milliseconds on client RPC side"
+  measurement_descriptor_name: "grpc.io/client/roundtrip_latency"
+  interval_aggregation: {
+    interval_sizes: {
+      seconds: 60
+      nanos: 0
```
sebright: I think `nanos: 0` is unnecessary, since it has the default value.

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,620 @@
...
+      seconds: 3600
+      nanos: 0
+    }
+  }
+  tag_keys: [
+    "method" # Tag value will be the name of client RPC method/service
+  ]
+}
+
+# Census defined RPC server IntervalViewDescriptors.
```
sebright: Same here, I think this should be "server interval ViewDescriptors."

.gitignore
```diff
@@ -1,2 +1,4 @@
 bazel-*
 target
+.idea
+census-proto.iml
```
songy23: Yep, will checkout changes on this file.

BUILD
```diff
@@ -33,3 +39,11 @@ java_proto_library(
     name = "stats_context-proto-java",
     deps = ["stats_context_proto"],
 )
+
+java_proto_library(
+    name = "view_descriptor_constants-proto-java",
+    deps = [
+        "stats_context_proto",
```
songy23: That's a typo, thx for catching that

stats/view_descriptor_constants.proto
```diff
@@ -0,0 +1,41 @@
...
+//     BASE="filename"  (for example, BASE="rpc_constants")
+//     cat $BASE.textpb | \
+//     protoc --encode=google.instrumentation.ViewDescriptorConstants \
+//     view_descriptor_constants.proto > $BASE.pb
+// - To convert a binary proto file back to a text-format proto file, use:
+//     BASE="filename"  (for example, BASE="rpc_constants")
+//     cat $BASE.pb | \
+//     protoc --decode=google.instrumentation.ViewDescriptorConstants \
+//     view_descriptor_constants.proto > $BASE.textpb
+message ViewDescriptorConstants {
```
songy23: Yes I suppose so, in the future we may want to  support pre-defined view constants for other services, and the ViewDescriptorConstants could be reused. Besides, if we don't have this data structure, the rpc_constants.textpb will need to be divided into a lot of separate files (in order to be parsed individually as separate MeasurementDescriptor and ViewDescriptor).

stats/rpc_constants.textpb
```diff
@@ -0,0 +1,620 @@
+# Constants for collecting specific RPC stats, including:
+#   - Measurement Descriptor, which describes a data point (measurement) type.
+#   - View Descriptor, which specifies an AggregationDescriptor and a set of tag keys.
+#     View Descriptor has one of the two kinds of Aggregation types associated with the View,
+#     Interval Aggregation Descriptor or Distribution Aggregation Descriptor.
+#     Aggregations are broken down by the unique set of matching tag values for each measurement.
```
songy23: Yes that's a better way to document, I'll update the comments here.

stats/rpc_constants.textpb
songy23: Nice catch :)

stats/rpc_constants.textpb
songy23: Fixed

stats/rpc_constants.textpb
songy23: Fixed

stats/rpc_constants.textpb
songy23: Sure, this should be align with "client distribution ViewDescriptors"

stats/rpc_constants.textpb
songy23: Removed

songy23:
Here's another example that we need to store both the text-format protos and proto binaries in our repo:
https://github.com/grpc/grpc/tree/master/test/core/census/data

stats/rpc_constants.textpb
sebright: I like the change to the comment at the top of this file, though I think that the comments about `MeasurementDescriptor` and `ViewDescriptor` that you moved to `view_descriptor_constants.proto` aren't necessary, since `MeasurementDescriptor` and `ViewDescriptor` are already documented in `census.proto`.

songy23: Yeah I got your point here, this could be a little redundant with what we have in census.proto. I'll update the comments here to be more concise.

stats/rpc_constants.textpb
songy23: Just found the answer elsewhere. In short server/server_elapsed_time and server/server_latency are the same for languages other than C++, and for C++ they are populated from different sources. I'll add a clarification here.

zhangkun83: If the difference is semantic, it makes sense to keep them and other languages should implement them as defined. If they are semantically the same but it's just C++ happens to have two implementations, they should be consolidated.

songy23: Here it should fall into the 2nd case, I'll talk with Dino to see if we want to consolidate them. Thx

PR #28, opened 2017-03-18
Add support for gradle to build the instrumentation-proto.

bogdandrutu:
* Configured travis to build with gradle.
* Disable all java warnings when build the generated files. 
* To build the proto files simply run $./gradlew clean build.


build.gradle
```diff
@@ -0,0 +1,68 @@
...
+    compile 'com.google.protobuf:protobuf-java:3.2.0'
+}
+
+protobuf {
+    protoc {
+        // The artifact spec for the Protobuf Compiler
+        artifact = 'com.google.protobuf:protoc:3.2.0'
+    }
+
+    generateProtoTasks {
```
zhangkun83: You can remove this block since you don't need to configure the tasks.

PR #33, opened 2017-08-30
Cleanup readme, license and build files.

bogdandrutu:
A lot of small fixes:
* Update license to use OpenCensus Authors;
* Update project name to be OpenCensus;
* Update project groupId to be io.opencensus (we may have to change it to io.opencensus.proto later);
* Rename BUILD -> BUILD.bazel to fix TODO from gradle.
* Update protoc to 3.4.0

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,16 +1,16 @@
-# Copyright 2016, Google Inc.
+# Copyright 2016-17, OpenCensus Authors
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
+#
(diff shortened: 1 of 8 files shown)


BUILD.bazel
```diff
@@ -0,0 +1,21 @@
+package(default_visibility = ["//visibility:public"])
```
songy23: Does this file need a copyright header?

bogdandrutu: grpc-java does not have, grpc (c++) has. I don't know, any suggestions?

songy23: I'd suggest to keep the header as before.

bogdandrutu: Done.

PR #34, opened 2017-09-05
Add Span data model for OpenCensus.

bogdandrutu:


trace/span.proto
```diff
@@ -0,0 +1,323 @@
...
+        // Indicates a received RPC message.
+        RECV = 2;
+      }
+
+      // Type of NetworkEvent. Indicates whether the RPC message was sent or
+      // received.
+      Type type = 2;
+
+      // An identifier for the message, which must be unique in this span.
+      uint64 message_id = 3;
```
adriancole: so this is required? seems like another random to create. why is this important? Maybe mention why as there is a cost to another random per event.  The assumption is that within the scope of a span ID, there is a chance for clash on timestamp? or is it more about lookup?

trace/span.proto
```diff
@@ -0,0 +1,323 @@
...
+
+      // The number of uncompressed bytes sent or received.
+      uint64 uncompressed_message_size = 4;
+
+      // The number of compressed bytes sent or received.
```
adriancole: definitely this should be optional, at least mentioned so

trace/span.proto
```diff
@@ -0,0 +1,323 @@
...
+      uint64 compressed_message_size = 5;
+    }
+
+    // A `TimeEvent` can contain either an `Annotation` object or a
+    // `NetworkEvent` object, but not both.
+    oneof value {
+      // One or more key:value pairs.
+      Annotation annotation = 2;
```
adriancole: "Text annotation with a set of attributes." as opposed to key/value pairs

trace/span.proto
```diff
@@ -0,0 +1,323 @@
...
+  // A collection of `TimeEvent`s. A `TimeEvent` is a time-stamped annotation
+  // on the span, consisting of either user-supplied key:value pairs, or
+  // details of an RPC message sent/received on the network.
+  message TimeEvents {
+    // A collection of `TimeEvent`s.
+    repeated TimeEvent time_event = 1;
+
+    // The number of dropped annotations in all the included time events.
+    // If the value is 0, then no annotations were dropped.
+    int32 dropped_annotations_count = 2;
```
adriancole: this looks maintenance causing, which might be ok. For example, if another type of event is added, you need a new field.

trace/span.proto
```diff
@@ -0,0 +1,323 @@
...
+  io.opencensus.trace.proto.Status status = 11;
+
+  // A highly recommended but not required flag that identifies when a trace
+  // crosses a process boundary. True when the parent_span belongs to the
+  // same process as the current span.
+  google.protobuf.BoolValue same_process_as_parent_span = 12;
+
+  // An optional number of child spans that were generated while this span
+  // was active. If set, allows implementation to detect missing child spans.
+  google.protobuf.UInt32Value child_span_count = 13;
```
adriancole: ok so no counter for dropped, rather a counter for not-dropped

trace/span.proto
```diff
@@ -0,0 +1,323 @@
...
+  // The inclued links.
+  Links links = 10;
+
+  // An optional final status for this span.
+  io.opencensus.trace.proto.Status status = 11;
+
+  // A highly recommended but not required flag that identifies when a trace
+  // crosses a process boundary. True when the parent_span belongs to the
+  // same process as the current span.
+  google.protobuf.BoolValue same_process_as_parent_span = 12;
```
adriancole: is this expected to be used (ex shared span zipkin v1 style?)

trace/span.proto
```diff
@@ -0,0 +1,323 @@
...
+  message StackFrames {
+    // Stack frames in this call stack.
+    repeated StackFrame frame = 1;
+    // The number of stack frames that were dropped because there
+    // were too many stack frames.
+    // If this value is 0, then no stack frames were dropped.
+    int32 dropped_frames_count = 2;
+  }
+
+  // Stack frames in this stack trace.
```
adriancole: the call stack which originated this span

trace/span.proto
```diff
@@ -0,0 +1,323 @@
...
+  // The hash ID is used to conserve network bandwidth for duplicate
+  // stack traces within a single trace.
+  //
+  // Often multiple spans will have identical stack traces.
+  // The first occurrence of a stack trace should contain both the
+  // `stackFrame` content and a value in `stackTraceHashId`.
+  //
+  // Subsequent spans within the same request can refer
+  // to that stack trace by only setting `stackTraceHashId`.
+  uint64 stack_trace_hash_id = 2;
```
adriancole: this is a bit confusing, are you suggesting a pattern like where there is no existing trace, but a new trace is created in a for loop with a counter?

trace/span.proto
```diff
@@ -0,0 +1,323 @@
...
+}
+
+// Binary module.
```
adriancole: maybe package or something, but binary might be fine, but better top-level desc might help

trace/span.proto
```diff
@@ -0,0 +1,323 @@
...
+  // For example: main binary, kernel modules, and dynamic libraries
+  // such as libc.so, sharedlib.so.
+  TruncatableString module = 1;
+
+  // A unique identifier for the module, usually a hash of its
+  // contents.
+  TruncatableString build_id = 2;
+}
+
+// Represents a string that might be shortened to a specified length.
```
adriancole: interesting as this hints that truncation will occur at origin.. or... a decorating proxy can also do this, too I suppose (just thinking)

trace/span.proto
```diff
@@ -0,0 +1,323 @@
...
+option java_multiple_files = true;
+option java_package = "io.opencensus.trace.proto";
+option java_outer_classname = "SpanProto";
+
+// A span represents a single operation within a trace. Spans can be
+// nested to form a trace tree. Often, a trace contains a root span
+// that describes the end-to-end latency, and one or more subspans for
+// its sub-operations. A trace can also contain multiple root spans,
+// or none at all. Spans do not need to be contiguous - there may be
+// gaps or overlaps between spans in a trace.
```
adriancole: an example of a span would be helpful (saying what it is, ex a grpc client send)

trace/span.proto
bogdandrutu: Thing about a stream of messages, RPC systems will usually have this as an incremented number.

trace/span.proto
bogdandrutu: Everything in this is optional unless it says required.

trace/span.proto
bogdandrutu: Good catch.

trace/span.proto
bogdandrutu: unfortunately proto does not have generics 👎

trace/span.proto
adriancole: maybe clarify "When present, must be unique, but needn't be random. For example, this could be a sequence id"

trace/span.proto
bogdandrutu: this is to replace the notion of span type (client/server/local). We think that in case of proxies for example you don't want to necessary create 2 spans to fit that model.

trace/span.proto
bogdandrutu: We are positive 👍

trace/span.proto
bogdandrutu: Added a TODO for the moment.

trace/span.proto
bogdandrutu: it can happen at any point (library, proxy, backend). The reason why we have it here is mainly because it can happen in the library.

trace/span.proto
bogdandrutu: Done.

trace/span.proto
bogdandrutu: No, think about an incoming request that generates 10 calls to ServiceA. For each call you create a Span and you want to record the StackTrace, then you will have all the 10 StackTraces probably the same.

trace/span.proto
bogdandrutu: ACK. I will ask for clarification (I am not the one who proposed this StackTrace format, we worked with an internal team that does pprof).

trace/span.proto
adriancole: ic gotcha

bogdandrutu:
I need an LGTM 👍

pom.xml
```diff
@@ -32,7 +32,8 @@
           <protocExecutable>protoc</protocExecutable>
           <protoSourceRoot>.</protoSourceRoot>
 	        <includes>
-	          <include>stats/*.proto</include>
+            <include>stats/*.proto</include>
```
dinooliva: nit: indent level

adriancole:
In zipkin would be nice to have local and remote endpoint details. We can chat later about this

PR #40, opened 2017-09-28
Rename NetworkEvents to MessageEvents

bogdandrutu:
Changes in this PR:
1) Rename NetworkEvents to MessageEvents;
2) Remove "message" from all the internal fields of the newly "MessageEvent" because the "message" is redundant.
3) Remove kernel timestamp for the moment because that is not generic enough to be used in MessageEvents (we can always add it later).

This is to allow us to use MessageEvents in pipeline works, in-process biderctional streams, or in batch processing and not force to have an RPC when we record them.

### DIFF
--- a/trace/trace.proto
+++ b/trace/trace.proto
@@ -105,38 +105,31 @@ message Span {
       Attributes attributes = 2;
     }
 
-    // An event describing an RPC message sent or received on the network.
-    message NetworkEvent {
-      // For sent messages, this is the time at which the first bit was sent.
-      // For received messages, this is the time at which the last bit was
-      // received.
-      google.protobuf.Timestamp time = 1;
-
-      // Indicates whether the network message was sent or received.
+    // An event describing a message sent/received between Spans.
+    message MessageEvent {
+      // Indicates whether the message was sent or received.
       enum Type {
         // Unknown event type.
         TYPE_UNSPECIFIED = 0;
-        // Indicates a sent RPC message.
+        // Indicates a sent message.
         SENT = 1;
-        // Indicates a received RPC message.
-        RECV = 2;
+        // Indicates a received message.
+        RECEIVED = 2;
       }
 
-      // Type of NetworkEvent. Indicates whether the RPC message was sent or
+      // Type of MessageEvent. Indicates whether the message was sent or
       // received.
-      Type type = 2;
+      Type type = 1;
 
-      // An identifier for the message, which must be unique in this span and
-      // the same between RECV and SENT (this allows to identify a message
-      // between the sender and receiver). For example, this could be a
-      // sequence id.
-      uint64 message_id = 3;
+      // An identifier for the message, which must be unique in this span.
+      uint64 id = 2;
 
       // The number of uncompressed bytes sent or received.
-      uint64 uncompressed_message_size = 4;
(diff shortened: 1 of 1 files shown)


trace/trace.proto
```diff
@@ -146,7 +139,7 @@ message Span {
       Annotation annotation = 2;
 
       // An event describing an RPC message sent/received on the network.
-      NetworkEvent network_event = 3;
+      MessageEvent network_event = 3;
```
adriancole: Ps you still want the old field name?

trace/trace.proto
```diff
@@ -105,38 +105,31 @@ message Span {
...
-      Type type = 2;
+      Type type = 1;
 
-      // An identifier for the message, which must be unique in this span and
-      // the same between RECV and SENT (this allows to identify a message
-      // between the sender and receiver). For example, this could be a
-      // sequence id.
-      uint64 message_id = 3;
+      // An identifier for the message, which must be unique in this span.
+      uint64 id = 2;
```
sebright: Can there be more than one MessageEvent per message, within one span?

trace/trace.proto
```diff
@@ -105,38 +105,31 @@ message Span {
...
 
       // The number of uncompressed bytes sent or received.
-      uint64 uncompressed_message_size = 4;
+      uint64 uncompressed_size = 3;
 
-      // The number of compressed bytes sent or received.
-      uint64 compressed_message_size = 5;
+      // The number of compressed bytes sent or received. If missing assumed to
+      // be the same size as uncompressed.
+      uint64 compressed_size = 4;
```
sebright: > If missing assumed to
     // be the same size as uncompressed.

Should the default be an impossible value, like -1?

trace/trace.proto
sebright: Several of the comments still mention NetworkEvent, too.

trace/trace.proto
bogdandrutu: Yes. Think about a stream request/response we record one entry for each part of the stream.

trace/trace.proto
bogdandrutu: Default is 0, it is impossible to send a 0 length message correct?

bogdandrutu:
Fixed the NetworkEvents before you sent your comments.

trace/trace.proto
sebright: I think it would help to explain that in the comment, like the comment on master.  The comment in the PR could be misread as saying that the MessageEvent's id must be unique, especially since MessageEvent is also a Protocol Buffer message.

trace/trace.proto
sebright: Probably, but I thought that -1 would be even clearer.

bogdandrutu: it is unsigned so no -1 that is max value.

sebright: Okay

trace/trace.proto
bogdandrutu: Can you be more explicit how do you want me to rephrase. Happy to change.

bogdandrutu: Good suggestion. Thanks.

bogdandrutu:
PTAL.

PR #49, opened 2018-02-05
Trace proto: replace end_time with duration; add float attr

ramonza:


opencensus/proto/trace/trace.proto
```diff
@@ -246,6 +251,8 @@ message AttributeValue {
     int64 int_value = 2;
     // A Boolean value represented by `true` or `false`.
     bool bool_value = 3;
+    // A 64-bit floating pooint value.
+    double float_value = 4;
```
bogdandrutu: We do not support this for the moment.

opencensus/proto/trace/trace.proto
```diff
@@ -62,10 +66,11 @@ message Span {
...
 
-  // The end time of the span. On the client side, this is the time kept by
-  // the local machine where the span execution ends. On the server side, this
-  // is the time when the server application handler stops running.
-  google.protobuf.Timestamp end_time = 6;
+  // The duration of the span. On the client side, this is the time kept by
+  // the local machine from start_time until the span execution ends. On the server
+  // side, this is the duration between the start_time and when the server
+  // application handler stops running.
+  google.protobuf.Duration duration = 14;
```
bogdandrutu: Why do we need this? end_time - start_time must be equal to this.

adriancole:
In zipkin we used duration not end time so that you can cheaply index
duration without math first. Also it discourages use of potentially
different clock times between start and finish which was a problem with
negative duration spans. Careful coding you can make start and finish work
by using only monotonic, but you still end up with math indexing duration
which is not that great. Another thing is that users view data (when
represented in json) it is easier to see what the duration was without math.

Not sure float should be used unless it is fixed offset. If fixed offset
sounds fine.

adriancole:
Oh another thing which is indirect here because census has an api. Metrics
can be harvested with duration easier than math even if math easy. Also log
statements based on harvested data are simply mapping fields no math. Ex
http logging format has optional duration as last field.

opencensus/proto/trace/trace.proto
ramonza: To avoid questions of wall clock changes https://github.com/census-instrumentation/opencensus-proto/pull/49#issuecomment-370283102

rakyll:
What's the status of this change? I would prefer us to have a consensus before we start to finalize APIs of some language implementations.

bogdandrutu:
If we replace end_time with duration we should also replace timestamps in the TimedEvents. I am not 100% convinced this change worth the effort to make this change everywhere. I hear a lot of arguments here about people looking at raw data, which I don't know is a common case, but I may be wrong.

Anyway I think if we really want to change to "durantion" everything must be a duration since the start_span.

ramonza:
what effort is "worth the effort"? where is this proto used that it's so hard to change? duration is smaller and doesn't raise the confusing issue of having to compute an end time in the timeline of the start time - an operation that is very hard to even explain to people and must be done in every language. i guess i just don't see what the cost of making this change now are maybe i'm missing something about where this proto is used / who is depending on it?

adriancole:
Fwiw I dont think everything has to be duration if span duration is finish.

Ex events dont have exactly the same problem unless you expect people to
make metrics by comparing event pairs... making events offset based is
possible and maybe good (I have considered doing this as well with zipkin
annotation just we dont do paired events much), anyway do think it is
separate issue.

Tldr span.duration is much quicker decision even if I agree we can think
through if event should als be offset.

bogdandrutu:
@ramonza the SpanData in all the languages (at least Java, Python) is consistent with the proto definition. Even if we do not use the proto directly this reflects our data model and the data we produce. If you want to avoid this problem that you mention of implementation recording wrong timestamps then this has to be modified as well.

@adriancole I kind of disagree with you regarding being different issues. I think consistency in a data-model is very important. Also I am hearing few reasons why this should be duration (avoid mistakes in the instrumentation libraries for example, this applies to all timestamps).

ramonza:
@bogdandrutu I am ok to delay this until we are PR-available for at least 1 lang

ramonza:
Old PR, closing

PR #50, opened 2018-02-07
Add generated Go code.

ramonza:


gen-go/statsproto/doc.go
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+// Package statsproto contains generated Protobuf code for OpenCensus stats.
+package statsproto
```
sebright: stats.proto hasn't been updated for the new design in https://github.com/census-instrumentation/opencensus-specs yet.

ramonza: Not sure what the new design is but let's fix that separately.

tsloughter:
I wouldn't this just live in the Go repo?

ramonza:
@tsloughter  https://github.com/census-instrumentation/opencensus-go/pull/406
not exactly sure why (I'm sure there's a good reason) but there's already code in this repo to support Java so I don't feel too bad adding Go as well.

gen-go/statsproto/doc.go
sebright: Some of the data types have completely changed, such as MeasurementDescriptor, DistributionAggregation, and IntervalAggregation.  I agree that it should be updated separately from this PR, though.

tsloughter:
Ah, ok. Instead of a submodule we just put it in the readme for the Erlang library, hehe https://github.com/census-instrumentation/opencensus-erlang#updating-opencensus-standard-protobuf-encoder-and-decoder

PR #51, opened 2018-03-03
Add Span kind to the Span.

bogdandrutu:
Proposal for adding a span kind that can be used also as an identity for the Span, this way we can remove the prefix "Sent." and "Recv." from the span names in grpc and http.

### DIFF
--- a/opencensus/proto/trace/trace.proto
+++ b/opencensus/proto/trace/trace.proto
@@ -31,6 +31,8 @@ option go_package = "traceproto";
 // its sub-operations. A trace can also contain multiple root spans,
 // or none at all. Spans do not need to be contiguous - there may be
 // gaps or overlaps between spans in a trace.
+//
+// The next id is 15.
 // TODO(bdrutu): Add an example.
 message Span {
   // A unique identifier for a trace. All spans from the same trace share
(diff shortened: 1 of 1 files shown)


opencensus/proto/trace/trace.proto
```diff
@@ -59,6 +61,26 @@ message Span {
...
+    // Unspecified.
+    SPAN_KIND_UNSPECIFIED = 0;
+
+    // Indicates that the span covers server-side handling of an RPC or other
+    // remote network request.
+    SERVER = 1;
+
+    // Indicates that the span covers the client-side wrapper around an RPC or
+    // other remote request.
+    CLIENT = 2;
```
adriancole: I would add PRODUCER and CONSUMER as they are widely used, and indicate sending to a broker instead of directly to a server and visa versa

HailongWen: OpenTracing also has [producer and consumer](https://github.com/opentracing/opentracing-java/blob/3e270abc6d07315ff71fe09886a9cbe93ea221f4/opentracing-api/src/main/java/io/opentracing/tag/Tags.java#L41) types. Might be good if we have that.

adriancole:
This is good progress, and CLIENT/SERVER is already handled in census by grpc instrumentation (and the upcoming http work). Let's not forget PRODUCER/CONSUMER but I agree that there should be some code driving that. In java you can look at this https://github.com/openzipkin/brave/tree/master/instrumentation/kafka-clients

and also the span kind definition in zipkin for later stuff. For now, let's merge as it gets a large percentage of instrumentation.. instrumentable.

PR #55, opened 2018-05-15
Metric: Move label keys to metric descriptor.

songy23:
Based on our discussion last week.

### DIFF
--- a/opencensus/proto/stats/metrics/metrics.proto
+++ b/opencensus/proto/stats/metrics/metrics.proto
@@ -17,6 +17,7 @@ syntax = "proto3";
 package opencensus.proto.stats.metrics;
 
 import "google/protobuf/timestamp.proto";
+import "google/protobuf/wrappers.proto";
 
 option go_package = "metricsproto";
 option java_multiple_files = true;
@@ -82,7 +83,7 @@ message TimeSeries {
   // The set of label values that uniquely identify this timeseries. Apply to all
   // points. The order of label values must match that of label keys in the
   // metric descriptor.
-  repeated string label_values = 1;
+  repeated google.protobuf.StringValue label_values = 1;
 
   // The data points of this timeseries. Point type MUST match the MetricDescriptor.type, so for
   // a CUMULATIVE type only cumulative_points are present and for a GAUGE type only gauge_points
--- a/opencensus/proto/stats/metrics/metrics.proto
+++ b/opencensus/proto/stats/metrics/metrics.proto
@@ -69,15 +69,20 @@ message MetricDescriptor {
   Type type = 4;
 
   // TODO: Consider to add LabelInfo{key, description} if needed.
+  // The label keys associated with the metric descriptor. Label keys must be
+  // unique.
+  repeated string label_keys = 5;
 }
 
 // A collection of data points that describes the time-varying values
 // of a metric.
 message TimeSeries {
   // TODO: Add restrictions for characters that can be used for keys and values.
+  // TODO: Wrap values to distinguish between empty-string and not-set.
   // The set of label values that uniquely identify this timeseries. Apply to all
-  // points.
-  map<string, string> label = 1;
+  // points. The order of label values must match that of label keys in the
+  // metric descriptor.
+  repeated string label_values = 1;
 
   // The data points of this timeseries. Point type MUST match the MetricDescriptor.type, so for
   // a CUMULATIVE type only cumulative_points are present and for a GAUGE type only gauge_points
--- a/opencensus/proto/stats/metrics/metrics.proto
+++ b/opencensus/proto/stats/metrics/metrics.proto
@@ -17,7 +17,6 @@ syntax = "proto3";
 package opencensus.proto.stats.metrics;
 
 import "google/protobuf/timestamp.proto";
-import "google/protobuf/wrappers.proto";
 
 option go_package = "metricsproto";
 option java_multiple_files = true;
@@ -82,7 +81,7 @@ message TimeSeries {
   // The set of label values that uniquely identify this timeseries. Apply to all
   // points. The order of label values must match that of label keys in the
   // metric descriptor.
-  repeated google.protobuf.StringValue label_values = 1;
+  repeated string label_values = 1;
 
   // The data points of this timeseries. Point type MUST match the MetricDescriptor.type, so for
   // a CUMULATIVE type only cumulative_points are present and for a GAUGE type only gauge_points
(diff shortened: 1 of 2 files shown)


opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -69,15 +69,20 @@ message MetricDescriptor {
   Type type = 4;
 
   // TODO: Consider to add LabelInfo{key, description} if needed.
+  // The label keys associated with the metric descriptor. Label keys must be
+  // unique.
+  repeated string label_keys = 5;
```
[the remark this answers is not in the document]
songy23: Good question, we already did this in Java ([PR/1108](https://github.com/census-instrumentation/opencensus-java/pull/1108)), though I'm not sure if we should enforce this in the proto description.

WDYT? @sebright @bogdandrutu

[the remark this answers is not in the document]
songy23: SGTM, updated

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -69,15 +69,20 @@ message MetricDescriptor {
...
 // of a metric.
 message TimeSeries {
   // TODO: Add restrictions for characters that can be used for keys and values.
+  // TODO: Wrap values to distinguish between empty-string and not-set.
   // The set of label values that uniquely identify this timeseries. Apply to all
-  // points.
-  map<string, string> label = 1;
+  // points. The order of label values must match that of label keys in the
+  // metric descriptor.
+  repeated string label_values = 1;
```
bogdandrutu: Use google.protobuf.StringValue see for example:
https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/trace/trace.proto#L247

You need to add `import "google/protobuf/wrappers.proto";`

songy23: Done.

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -69,15 +70,20 @@ message MetricDescriptor {
...
+  // The label keys associated with the metric descriptor. Label keys must be
+  // unique and in sorted order.
+  repeated string label_keys = 5;
 }
 
 // A collection of data points that describes the time-varying values
 // of a metric.
 message TimeSeries {
   // TODO: Add restrictions for characters that can be used for keys and values.
+  // TODO: Wrap values to distinguish between empty-string and not-set.
```
bogdandrutu: Remove this TODO?

songy23: Done.

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -69,15 +70,19 @@ message MetricDescriptor {
   Type type = 4;
 
   // TODO: Consider to add LabelInfo{key, description} if needed.
+  // The label keys associated with the metric descriptor. Label keys must be
+  // unique and in sorted order.
```
g-easy: As discussed offline, we don't want to do too much validation. I think the order of keys should be left unspecified.

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -69,15 +70,19 @@ message MetricDescriptor {
...
 // A collection of data points that describes the time-varying values
 // of a metric.
 message TimeSeries {
   // TODO: Add restrictions for characters that can be used for keys and values.
   // The set of label values that uniquely identify this timeseries. Apply to all
-  // points.
-  map<string, string> label = 1;
+  // points. The order of label values must match that of label keys in the
+  // metric descriptor.
+  repeated google.protobuf.StringValue label_values = 1;
```
g-easy:
```
// Wrapper message for `string`.
//
// The JSON representation for `StringValue` is JSON string.
message StringValue {
  // The string value.
  string value = 1;
}
```

How does this let us distinguish between empty string and unset?

opencensus/proto/stats/metrics/metrics.proto
songy23: That's right, removed. I think we don't want to validate if the tag keys are unique either.

opencensus/proto/stats/metrics/metrics.proto
songy23: IIUC for empty string we explicitly set `""` for the `value` field of `StringValue`, while for unset we don't set the `value` field (i.e use a default instance).

/cc @bogdandrutu

bogdandrutu: label_value will be null (the whole message) not the string value inside the message.

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -69,15 +70,18 @@ message MetricDescriptor {
   Type type = 4;
 
   // TODO: Consider to add LabelInfo{key, description} if needed.
+  // The label keys associated with the metric descriptor.
+  repeated string label_keys = 5;
 }
```
dinooliva: Do we want to specify an order? Probably not given our 'passthrough' philosophy but it is an issue we've discussed before.

songy23: I thought the decision is not to do any validation here?

From @g-easy : 
> We want to avoid performing validation as much as we can when dealing with Metrics.

bogdandrutu: No ordering please.

opencensus/proto/stats/metrics/metrics.proto
bogdandrutu: I see, I am not 100% convinced that this whole optimization makes sense with all these complicated things. But here is how I think we can solve this issue:

message LabelValueOrNull {
  // The value for the label.
  string value = 1;
  // If false the value field is ignored and considered not set.
  bool has_value = 2; 
}

opencensus/proto/stats/metrics/metrics.proto
dinooliva: sgtm

opencensus/proto/stats/metrics/metrics.proto
songy23: I think that could go into a separate PR - I put back the TODO about differentiating between empty and unset.

g-easy: +1 for separate PR. Do you want to change `StringValue` back to a regular `string` and drop the `wrappers_proto` dependency?

songy23:
Thanks!

PR #56, opened 2018-05-18
Metric: Wrap label values to distinguish between empty-string and unset.

songy23:
We need a way to distinguish between empty-string and unset for label values. Since `label_values` is a repeated field, using the built-in `Wrapper.StringValue` doesn't help in this case. Instead we need to define our own `LabelValue` type that wraps the string value, as suggested by @bogdandrutu in https://github.com/census-instrumentation/opencensus-proto/pull/55#discussion_r189125133.


songy23:
Will merge this PR if there are no more comments/concerns by EOD tomorrow (PST).

songy23:
Thanks all!

PR #57, opened 2018-05-22
Metric: Add LabelKey type that has key and description.

songy23:


opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -68,9 +68,17 @@ message MetricDescriptor {
   }
   Type type = 4;
 
-  // TODO: Consider to add LabelInfo{key, description} if needed.
```
g-easy: Why is description needed?

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -68,9 +68,17 @@ message MetricDescriptor {
...
-  repeated string label_keys = 5;
+  repeated LabelKey label_keys = 5;
+}
+
+// Defines a label key associated with a metric descriptor.
+message LabelKey {
+  // The key for the label.
+  string key = 1;
+
+  // The description of this label key.
```
g-easy: Please elaborate. What is the description for? Do any current exporters use this, or is it just a comment for human readers?

songy23: It's an optional field for other monitoring backend (like Stackdriver Monitoring). For now our exporters need to hard-code the description field: [CPP](https://github.com/census-instrumentation/opencensus-cpp/blob/master/opencensus/exporters/stats/stackdriver/internal/stackdriver_utils.cc#L40), [Java](https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java#L79).

In addition, I think adding description for label key could help human reader, especially for keys like rpc_client_method vs rpc_server_method. We also have descriptions in the [specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md#tags).

opencensus/proto/stats/metrics/metrics.proto
g-easy: Answer: some backends make use of it?

opencensus/proto/stats/metrics/metrics.proto
g-easy: How about something like "a human-readable description of what this labelkey represents"

opencensus/proto/stats/metrics/metrics.proto
songy23: Yes, 1. it's useful for some backend, and 2. provides more information to human readers.

opencensus/proto/stats/metrics/metrics.proto
songy23: Sounds good, done.

PR #60, opened 2018-05-24
Add export service proto

rakyll:
We are experimentally working on a an export service
that will allow OpenCensus libraries to report collected
data to an external service for processing and exporting.

Some frameworks and ecosystems are now providing out-of-the-box
instrumentation by using OpenCensus but the user is still expected
to register an exporter in order to export data. This is a
problem during an incident. Even though our users can benefit
from having more diagnostics data coming out of services
already instrumented with OpenCensus, they have to modify
their code to register an exporter and redeploy. Asking our
users recompile and redeploy is not an ideal at an incident
time.

This experiment might be a possible solution to
eliminate this requirement.

Updates https://github.com/census-instrumentation/opencensus-specs/issues/72.

### DIFF
--- a/opencensus/proto/export/BUILD.bazel
+++ b/opencensus/proto/export/BUILD.bazel
@@ -1,0 +1,34 @@
+# Copyright 2018, OpenCensus Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//visibility:public"])
+
+proto_library(
+    name = "export_proto",
+    srcs = ["export.proto"],
+    deps = [
+        "@com_google_protobuf//:duration_proto",
+        "@com_google_protobuf//:timestamp_proto",
+    ],
+)
+
+cc_proto_library(
+    name = "export_proto_cc",
+    deps = [":export_proto"],
+)
+
+java_proto_library(
+    name = "export_proto_java",
+    deps = [":export_proto"],
+)
--- a/opencensus/proto/export/BUILD.bazel
+++ b/opencensus/proto/export/BUILD.bazel
@@ -18,8 +18,8 @@ proto_library(
     name = "export_proto",
     srcs = ["export.proto"],
     deps = [
-        "@com_google_protobuf//:duration_proto",
-        "@com_google_protobuf//:timestamp_proto",
+        "//opencensus/proto/trace:trace_proto",
+        "//opencensus/proto/stats/metrics:metrics_proto"
     ],
 )
 
--- a/opencensus/proto/export/export.proto
+++ b/opencensus/proto/export/export.proto
@@ -25,7 +25,7 @@ package export;
 import "trace/trace.proto";
 import "stats/metrics/metrics.proto";
 
-service ExportService {
+service Export {
   rpc Export (ExportRequest) returns (ExportResponse) {}
 }
 
--- a/opencensus/proto/export/export.proto
+++ b/opencensus/proto/export/export.proto
@@ -22,8 +22,8 @@ option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/e
 
 package export;
 
-import "trace/trace.proto";
-import "stats/metrics/metrics.proto";
+import "opencensus/proto/trace/trace.proto";
+import "opencensus/proto/stats/metrics/metrics.proto";
 
 service Export {
   rpc Export (ExportRequest) returns (ExportResponse) {}
--- a/opencensus/proto/export/export.proto
+++ b/opencensus/proto/export/export.proto
@@ -14,16 +14,16 @@
 
 syntax = "proto3";
 
+package opencensus.proto.export;
+
+import "opencensus/proto/trace/trace.proto";
+import "opencensus/proto/stats/metrics/metrics.proto";
+
 option java_multiple_files = true;
 option java_package = "io.opencensus.export";
 option java_outer_classname = "ExportProto";
 
 option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/exportproto";
-
-package export;
-
-import "opencensus/proto/trace/trace.proto";
-import "opencensus/proto/stats/metrics/metrics.proto";
 
 service Export {
   rpc Export (ExportRequest) returns (ExportResponse) {}
--- a/opencensus/proto/export/export.proto
+++ b/opencensus/proto/export/export.proto
@@ -1,0 +1,39 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


opencensus/proto/export/export.proto
```diff
@@ -0,0 +1,39 @@
...
+
+import "trace/trace.proto";
+import "stats/metrics/metrics.proto";
+
+service ExportService {
+  rpc Export (ExportRequest) returns (ExportResponse) {}
+}
+
+message ExportRequest {
+  repeated opencensus.proto.trace.Span spans = 1;
```
bogdandrutu: I think for consistency we have two options:
1) add a SpanSet similar to MetricSet
2) use repeated Metric

I would prefer 1).

rakyll: I went with the option 2 because it is not quite clear when we will use SpanSet elsewhere at this moment. PTAL.

opencensus/proto/export/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//visibility:public"])
+
+proto_library(
+    name = "export_proto",
+    srcs = ["export.proto"],
+    deps = [
+        "@com_google_protobuf//:duration_proto",
```
bogdandrutu: You don't need this correct? same for timestamp.

You should add dependency on the metric_proto and span_proto.

opencensus/proto/export/export.proto
```diff
@@ -0,0 +1,39 @@
...
+import "stats/metrics/metrics.proto";
+
+service Export {
+  rpc Export (ExportRequest) returns (ExportResponse) {}
+}
+
+message ExportRequest {
+  repeated opencensus.proto.trace.Span spans = 1;
+
+  repeated opencensus.proto.stats.metrics.Metric metrics = 2;
```
ramonza: Why not have separate RPCs or even separate services for spans and metrics? This would allow us more visibility into the two quite distinct operations.

opencensus/proto/export/export.proto
ramonza: Why use inline batching here instead of using a streaming RPC? This seems like an ideal use for streaming RPCs because you could avoid having to allocate a large batch of Span objects and you also don't have to manually batch.

opencensus/proto/export/BUILD.bazel
rakyll: Done.

opencensus/proto/export/export.proto
rakyll: The method takes repeated Spans because we wanted to give the clients the ability to batch before putting things on wire. Otherwise, we can avoid having a repeated fields.

opencensus/proto/export/export.proto
rakyll: I haven't separated them for the connivence of the exporter implementations. I expect the exporter to batch some metrics and spans and upload once a while. If we want to have more visibility into metrics vs spans calls, we should split.

ramonza: I don't really see how this is more convenient for exporter implementations than having separate RPCs. Maybe I am not imagining the same implementation but today the exporters are pretty different in how they batch (stats vs. trace). And in any case I would say that the observability advantages of having separate RPCs or even separate services are more important.

opencensus/proto/export/export.proto
ramonza: we may want to rationalize this with https://github.com/envoyproxy/data-plane-api/blob/1dcdaaaf0853c5258aa9d1588a052ca51dbbfa7d/envoy/service/trace/v2/trace_service.proto

I like their approach of using both a repeated field and streaming RPC - that way the client gets maximum freedom to optimize it's batching and memory usage. With a streaming request, I'm not sure there is any reason to do batching of objects inside the application rather than just sending them as soon as they are available and allowing the batching to occur lower down the stack.

ramonza:
I am not sure this repo is the right place for this at this time. I think it might be better to work up a prototype first and collocate the proto with the prototype code to facilitate rapid iteration. I don't think it helps to have the proto definition here in the "official" proto repo until it is relatively finalized.

bogdandrutu:
@ramonza having repeated still allows streaming/batching so no difference on how you implement it. Actually I feel this gives us more freedom to do what we want.

@ramonza I am fine either way (having it here or in a separate repo).

rakyll:
Having here is making it easier to import the existing protos without a lot of hassle.

rakyll:
Fixed the build, please take another look.

ramonza:
Batching using repeated (rather than streaming) forces you to construct the entire batch request object in memory, effectively doubly your memory usage. If you used streaming, you could re-use the same message object for each item, translating it to proto "just in time" to send it over the wire. I really don't understand the motivation for not using streaming here, it seems like the perfect use case for it.

adriancole:
autocorrect... last I meant to write was "re-couple, right?"

opencensus/proto/exporter/exporter.proto
```diff
@@ -0,0 +1,39 @@
...
+import "opencensus/proto/stats/metrics/metrics.proto";
+
+option java_multiple_files = true;
+option java_package = "io.opencensus.proto.exporter";
+option java_outer_classname = "ExporterProto";
+
+option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/exporterproto";
+
+service Export {
+  rpc Export (ExportRequest) returns (ExportResponse) {}
```
bogdandrutu: To not have to start an RPC every request I would suggest to use bi-directional streams:
rpc Export (stream ExportRequest) returns (stream ExportResponse) {}

adriancole:
I think separate services is best honestly, as those doing trace export
don't need to be bothered if there are changes in the metrics proto (and
any future protos like exception reporting, logging or whatever). Meanwhile
cohosting multiple services I'm assuming isn't a big deal. If I'm naive
about this, please let me know as I don't currently run gRPC services.

bogdandrutu:
@adriancole I don't think it is a problem to have 2 different services. I am happy with the proposal to have two services.

rakyll:
Thanks for the feedback. I updated the proto accordingly.

I kept the export messages with repeated fields but the services are streaming at this point according to the last feedback from @bogdandrutu. I am not sure if it is required/useful.

One more thing to note, this proto is not finalized yet. The only user of it will be the prototype exporter daemon mentioned in the commit message. We are expecting to change it in a backwards compatible way in a few weeks given no one else is depending on it. I need this proto to be merged to work on the initial commit of the exporter daemon. PTAL.

adriancole:
maybe old fashioned, but could help to add a comment towards the
experimental status in the proto? not sure if this is done usually, but
could help relay in future changes to not worry so much about comments wrt
back compat or otherwise.

bogdandrutu:
@adriancole I completely agree.

opencensus/proto/exporter/exporter.proto
```diff
@@ -0,0 +1,48 @@
...
+// NOTE: This proto is experimental and is subject to change at this point.
+// Please do not use it at the moment.
+
+package opencensus.proto.exporter;
+
+import "opencensus/proto/trace/trace.proto";
+import "opencensus/proto/stats/metrics/metrics.proto";
+
+option java_multiple_files = true;
+option java_package = "io.opencensus.proto.exporter";
```
songy23: Suggestion: rename the package to "export" instead of "exporter", since in Java "exporter" points to the actual exporters (Stackdriver, Prometheus).

opencensus/proto/exporter/exporter.proto
```diff
@@ -0,0 +1,48 @@
...
+option java_outer_classname = "ExporterProto";
```
songy23: Same here, better to rename it to "ExportProto"

opencensus/proto/exporter/exporter.proto
rakyll: What about proto exporter or something more specific?

songy23: Let's leave it as it is, for consistency across language :)

PR #62, opened 2018-06-12
Metrics: Add Exemplar to DistributionValue.

songy23:
/cc @bogdandrutu

### DIFF
--- a/opencensus/proto/stats/metrics/metrics.proto
+++ b/opencensus/proto/stats/metrics/metrics.proto
@@ -226,7 +226,8 @@ message DistributionValue {
   // Distribution values. They are metadata that gives information about a
   // particular value added to a Distribution bucket.
   message Exemplar {
-    // Value of the exemplar point.
+    // Value of the exemplar point. It determines which bucket the exemplar
+    // belongs to.
     double value = 1;
 
     // The observation (sampling) time of the above value.


opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -215,12 +215,30 @@ message DistributionValue {
...
+  // particular value added to a Distribution bucket.
+  message Exemplar {
+    // Value of the exemplar point.
+    int64 value = 1;
+
+    // The observation (sampling) time of the above value.
+    google.protobuf.Timestamp timestamp = 2;
+
+    // The trace id associated with the example value.
+    bytes trace_id = 3;
```
ramonza: What is the reason for supporting trace_id/span_id instead of arbitrary attachments like:  https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#Exemplar

I think there's lots of value in having other data associated with exemplars, for example the full tag set or maybe a stack trace.

songy23: In my opinion the advantage of having strong type instead of any is 1) we don't need to worry about encoding/decoding here, and 2) we can be sure that only a single attachment of any given message type is in a single exemplar, which is enforced by Stackdriver backend. If we want to have other attachments, we can add additional fields to Exemplar.

BTW Stackdriver doesn't really support arbitrary attachments in Exemplar. For now if we try to attach some message other than span context or dropped label to Exemplar, Stackdriver backend will return an error.

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -215,12 +215,30 @@ message DistributionValue {
...
   // If there is a histogram, then the sum of the values in the Bucket counts
   // must equal the value in the count field of the distribution.
   repeated Bucket buckets = 6;
+
+  // Exemplars are example points that may be used to annotate aggregated
+  // Distribution values. They are metadata that gives information about a
+  // particular value added to a Distribution bucket.
+  message Exemplar {
+    // Value of the exemplar point.
+    int64 value = 1;
```
ramonza: Shouldn't this be a double?

opencensus/proto/stats/metrics/metrics.proto
ramonza: The advantage of Any is that the backend can just store the items without actually interpreting them and return them in the API that the application can interpret any way it likes. The fact that Stackdriver enforces a particular kind seems to cut off useful applications unnecessarily. 

Regarding the validation of only a single type per Exemplar, I thought that we had decided that we would not try to do validation in the metrics proto? 

As an alternative to Any, What about changing this to be a `oneof` of SpanContext and dropped tags?

opencensus/proto/stats/metrics/metrics.proto
songy23: I think this `value` is just the bucket count so it should also be int64.

opencensus/proto/stats/metrics/metrics.proto
songy23: Using `oneof` sounds better, I updated the attachment proto. 

I agree that we should avoid doing validations, though I think it makes sense to have restriction on the type of attachments we want to Exemplar.

opencensus/proto/stats/metrics/metrics.proto
songy23: Took a second thought, sorry I got this wrong at first. `value` should be the actual value point instead of bucket count, so it should be of type `double`.

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -215,12 +215,49 @@ message DistributionValue {
...
+    double value = 1;
+
+    // The observation (sampling) time of the above value.
+    google.protobuf.Timestamp timestamp = 2;
+
+    // The context of a span to be attached to Exemplar.
+    message SpanContext {
+
+      // The resource name of the span.
+      string span_name = 1;
```
g-easy: I think this should be trace ID, not span_name.

How about:
```
// Unique identifier for a trace showing this exemplar.
// The ID is a 16-byte array.
bytes trace_id = 1;
```

Note that https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#Exemplar says attachments[] can be "https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#Exemplar"

And https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v1#google.devtools.cloudtrace.v1.Trace says trace_id is "128-bit numeric value formatted as a 32-byte hex string"

Whereas https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/trace/trace.proto has "bytes trace_id" documented to be a "16-byte array"

g-easy: And if we're linking to a span, then there should also be an 8-byte span_id. Otherwise rename this message from SpanContext to TraceSomething.

songy23: Make sense to me, and yes we need both the trace id and span id in span context.

Please note that the Exemplar attachment we're going to use for trace is the [span name from cloud trace v2](https://cloud.google.com/trace/docs/reference/v2/rest/v2/projects.traces/batchWrite#Span), instead of the [v1 trace_id](https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v1#google.devtools.cloudtrace.v1.Trace):

>The resource name of the span in the following format:

> projects/[PROJECT_ID]/traces/[TRACE_ID]/spans/[SPAN_ID]

>[TRACE_ID] is a unique identifier for a trace within a project; it is a 32-character hexadecimal encoding of a 16-byte array.

>[SPAN_ID] is a unique identifier for a span within a trace; it is a 16-character hexadecimal encoding of an 8-byte array.

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -235,8 +235,13 @@ message DistributionValue {
     // The context of a span to be attached to Exemplar.
     message SpanContext {
 
-      // The resource name of the span.
-      string span_name = 1;
+      // A unique identifier for a trace showing this exemplar.
+      // It is a 32-character hexadecimal encoding of a 16-byte array.
```
g-easy: Do you think this should be a binary (not hex) encoding in order to match OpenCensus's trace.proto?

The Stackdriver exporter will have to format it into the `projects/%s/traces/%s/etc` format like you said.

songy23: Yeah I think it makes more sense to avoid mentioning hex encoding in this proto. Updated the description here.

And yes, Stackdriver exporter needs to take care of the encoding and formatting (and only the Stackdriver exporter knows the project id to use).

songy23:
PTAL. Per our meeting today, we would like to have more generic interface for Exemplar attachment. Updated the attachment to be of type `Any` instead of concrete types. It's up to the exporters how to deal with the attachments.

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -232,35 +233,8 @@ message DistributionValue {
...
     // Contextual information about the example value.
-    oneof attachment {
-
-      // A span context attachment.
-      SpanContext span_context = 3;
-
-      // A dropped labels attachment.
-      DroppedLabels dropped_labels = 4;
-    }
+    repeated google.protobuf.Any attachments = 3;
```
sebright: Why does this field use Any, instead of a string to string map, as in https://github.com/census-instrumentation/opencensus-specs/pull/122?

songy23: Done. Cloud Monitoring uses Any as Exemplar attachment in their proto definition, though it makes our life easier to use a map on OpenCensus side.

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -215,12 +216,27 @@ message DistributionValue {
...
   // If there is a histogram, then the sum of the values in the Bucket counts
   // must equal the value in the count field of the distribution.
   repeated Bucket buckets = 6;
+
+  // Exemplars are example points that may be used to annotate aggregated
+  // Distribution values. They are metadata that gives information about a
+  // particular value added to a Distribution bucket.
+  message Exemplar {
+    // Value of the exemplar point.
+    double value = 1;
```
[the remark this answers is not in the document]
songy23: Yes the exemplar value field is required because it determines which bucket the exemplar belongs to, while things in 'attachments' are optional.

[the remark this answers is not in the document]
songy23: > Is the value useful for reasons other than associating it with the bucket (I would expect so)?

I think that depends on the backend. For Stackdriver, they didn't mention anything specific except that [value is used to determine to which bucket the exemplar belongs.](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#exemplar).

> Also, should Exemplar be a component of Bucket?

We can do that. I added Exemplar as a separate repeated field because this matches the [proto definition on Stackdriver](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#distribution). In our definition Bucket can have more information than just bucket count, and that's different from Stackdriver.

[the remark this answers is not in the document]
songy23: Reverted the change on having Exemplar as a part of Bucket, per our discussion today.

songy23:
PTAL, I changed back to use Map<String, String> for Exemplar.attachment, and keep the exemplar list separate from the bucket list.

opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -215,12 +215,28 @@ message DistributionValue {
...
+  message Exemplar {
+    // Value of the exemplar point. It determines which bucket the exemplar
+    // belongs to.
+    double value = 1;
+
+    // The observation (sampling) time of the above value.
+    google.protobuf.Timestamp timestamp = 2;
+
+    // Contextual information about the example value.
+    map<string, string> attachments = 3;
```
dinooliva: I'm still not sure about the name 'attachments' but I can't think of a better name offhand. On the up-side, we didn't go with 'context' :)

'annotations' is possible but  I don't want to confuse things with tracing.

songy23: I chose this name "attachments" to be consistent with [Stackdriver](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#exemplar) and Streamz. I think it's good to start with unless we came up with a better alternative.

dinooliva: Understood - I think being consistent w/Stackdriver and Monarch is compelling.

PR #63, opened 2018-06-14
Make MetricDescriptor.Type capture the type of the value as well.

g-easy:
e.g. int64 vs double vs distribution.

### DIFF
--- a/opencensus/proto/stats/metrics/metrics.proto
+++ b/opencensus/proto/stats/metrics/metrics.proto
@@ -64,7 +64,7 @@ message MetricDescriptor {
   //
   // A gauge is an instantaneous measurement of a value.
   //
-  // A cumulative measurements is a value accumulated over a time interval. In
+  // A cumulative measurement is a value accumulated over a time interval. In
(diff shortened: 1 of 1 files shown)


opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -61,18 +61,31 @@ message MetricDescriptor {
   string unit = 3;
 
   // The kind of metric. It describes how the data is reported.
+  //
+  // A gauge is an instantaneous measurement of a value.
+  //
+  // A cumulative measurement is a value accumulated over a time interval. In
+  // a time series, cumulative measurements should have the same start time and
+  // increasing end times, until an event resets the cumulative value to zero
```
songy23: nit: there are int64, double and distribution cumulative values, and "resets to zero" probably doesn't apply to distribution values. maybe remove "to zero"?

g-easy: Distributions do get reset to zero, e.g. on program start.

PR #65, opened 2018-06-14
Gauge vs Cumulative.

g-easy:
- Split TimeSeries into GaugeTimeSeries vs CumulativeTimeSeries.

- Merge GaugePoint and CumulativePoint into a single Point message
  to minimize validation.

### DIFF
--- a/opencensus/proto/stats/metrics/metrics.proto
+++ b/opencensus/proto/stats/metrics/metrics.proto
@@ -94,9 +94,8 @@ message LabelKey {
 // A collection of data points that describes the time-varying values
 // of a gauge metric.
 message GaugeTimeSeries {
-  // TODO: Add restrictions for characters that can be used for keys and values.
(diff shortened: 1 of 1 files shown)


opencensus/proto/stats/metrics/metrics.proto
```diff
@@ -90,31 +92,45 @@ message LabelKey {
...
+  repeated Point points = 2;
+}
+
+// A collection of data points that describes the time-varying values
+// of a cumulative metric.
+message CumulativeTimeSeries {
+  // The time that the cumulative value was reset to zero.
+  google.protobuf.Timestamp start_time = 1; // required
+
+  // TODO: Add restrictions for characters that can be used for keys and values.
```
songy23: IIRC we can remove this TODO now, since the decision is not to do any validation.

g-easy: Done.

PR #67, opened 2018-07-19
Remove Range from Distribution. No backend supports this.

bogdandrutu:
Reasons:
* No backend supports this. See comment for Stackdriver https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java#L379
* Usually in the UI data from cumulative distributions are "rated" (display a rate over "deltas"). This information cannot be preserved when computing deltas. It can only be useful if we support deltas but that is not the case.
* Simplify the Distribution metadata.

### DIFF
--- a/opencensus/proto/stats/metrics/metrics.proto
+++ b/opencensus/proto/stats/metrics/metrics.proto
@@ -180,19 +180,6 @@ message DistributionValue {
   // If count is zero then this field must be zero.
   double sum_of_squared_deviation = 3;
 
-  // The range of the population values.
-  message Range {
-    // The minimum of the population values.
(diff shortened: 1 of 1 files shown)


songy23:
It's still in the public proto too: https://github.com/googleapis/googleapis/blob/master/google/api/distribution.proto#L167

bogdandrutu:
I cannot speak for Stackdriver but I guess that they cannot support it because they cannot compute deltas for these values so they cannot calculate rates for example which is the most common thing when you have a cumulative value.

PR #70, opened 2018-07-30
Move exemplars in the Bucket.

bogdandrutu:
Reasons:

1. Validation of the ordering is required in the previous model so no extra validation required.
2. Small size improvement in the representation: If you have 2 exemplars then you need memory for the 2 exemplars object + extra memory for the repeated field (size, etc.). One null object (in the new representation) is 0 bytes on the wire.

### DIFF
--- a/opencensus/proto/metrics/metrics.proto
+++ b/opencensus/proto/metrics/metrics.proto
@@ -207,6 +207,9 @@ message DistributionValue {
     // The number of values in each bucket of the histogram, as described in
     // bucket_bounds.
     int64 count = 1;
+
+    // If the distribution does not have a histogram, then omit this field.
+    Exemplar exemplars = 2;
   }
 
   // If the distribution does not have a histogram, then omit this field.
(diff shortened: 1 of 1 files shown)


songy23:
I don't have specific objections to this PR. This is actually our initial proposal, but later we had a discussion and decided to keep `Exemplar` a separate list because (IIRC):

1. This matches the proto definition of Google API:
```proto
message Distribution {
...
repeated int64 bucket_counts = 7;
...
repeated Exemplar exemplars = 10;
}
```
2. Compatible with backend that supports >1 `Exemplar`s per bucket;
3. We don't do validation on the `Exemplar` list - it's up to the exporter. For
> Validation of the ordering is required in the previous model so no extra validation required.
4. Having a list instead of a field in another message is probably more efficient in encoding/decoding.

Maybe we should bring this up again in the cross-language sync.

opencensus/proto/metrics/metrics.proto
```diff
@@ -207,6 +207,9 @@ message DistributionValue {
     // The number of values in each bucket of the histogram, as described in
     // bucket_bounds.
     int64 count = 1;
+
+    // If the distribution does not have a histogram, then omit this field.
+    Exemplar exemplars = 2;
```
g-easy: No longer plural.
`Exemplar exemplar = 2;`

bogdandrutu: Done.

PR #71, opened 2018-07-30
Change from multiple types of timeseries to have one.

bogdandrutu:
Pros:

1. Easier to add new types (not gauge, or cumulative). This may not be immediately useful but in the future may make a difference.
2. Simplify the code by removing one extra type.
3. Compatible with SD and OpenMetrics effort.

Cons:

1. Checking that data are valid is harder in the new proposal. But OpenCensus tries to do minimal validation of the data (relying more on the apis/implementations that produce the data to do the right thing).

### DIFF
--- a/opencensus/proto/metrics/metrics.proto
+++ b/opencensus/proto/metrics/metrics.proto
@@ -119,9 +119,10 @@ message LabelValue {
 
 // A timestamped measurement.
 message Point {
-  // Available only for cumulative metrics. The time when the cumulative value
+  // Must be present for cumulative metrics. The time when the cumulative value
   // was reset to zero. The cumulative value is over the time interval
-  // [start_timestamp, timestamp].
+  // [start_timestamp, timestamp]. If not specified, the backend can use the
+  // previous recorded value.
   google.protobuf.Timestamp start_timestamp = 1;
 
   // The moment when this point was recorded. If not specified, the timestamp
--- a/opencensus/proto/metrics/metrics.proto
+++ b/opencensus/proto/metrics/metrics.proto
@@ -99,6 +99,12 @@ message LabelKey {
 // A collection of data points that describes the time-varying values
 // of a metric.
 message TimeSeries {
+  // Must be present for cumulative metrics. The time when the cumulative value
+  // was reset to zero. The cumulative value is over the time interval
+  // [start_timestamp, timestamp]. If not specified, the backend can use the
+  // previous recorded value.
+  google.protobuf.Timestamp start_timestamp = 1;
+
   // The set of label values that uniquely identify this timeseries. Applies to
   // all points. The order of label values must match that of label keys in the
   // metric descriptor.
@@ -119,26 +125,20 @@ message LabelValue {
 
 // A timestamped measurement.
 message Point {
-  // Must be present for cumulative metrics. The time when the cumulative value
-  // was reset to zero. The cumulative value is over the time interval
-  // [start_timestamp, timestamp]. If not specified, the backend can use the
-  // previous recorded value.
-  google.protobuf.Timestamp start_timestamp = 1;
-
   // The moment when this point was recorded. If not specified, the timestamp
   // will be decided by the backend.
-  google.protobuf.Timestamp timestamp = 2;
+  google.protobuf.Timestamp timestamp = 1;
 
   // The actual point value.
   oneof value {
     // A 64-bit integer.
-    int64 int64_value = 3;
+    int64 int64_value = 2;
 
     // A 64-bit double-precision floating-point number.
-    double double_value = 4;
+    double double_value = 3;
 
     // A distribution value.
-    DistributionValue distribution_value = 5;
+    DistributionValue distribution_value = 4;
 
     // TODO: Add support for Summary type. This is an aggregation that produces
     // percentiles directly.
--- a/opencensus/proto/metrics/metrics.proto
+++ b/opencensus/proto/metrics/metrics.proto
@@ -38,10 +38,8 @@ message Metric {
   MetricDescriptor metric_descriptor = 1;
 
   // One or more timeseries for a single metric, where each timeseries has
-  // one or more points. The type of the timeseries must match
-  // metric_descriptor.type, so only one of the two should be populated.
-  repeated GaugeTimeSeries gauge_timeseries = 2;
-  repeated CumulativeTimeSeries cumulative_timeseries = 3;
+  // one or more points.
+  repeated TimeSeries timeseries = 2;
 }
 
 // Defines a metric type and its schema.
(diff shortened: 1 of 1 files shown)


opencensus/proto/metrics/metrics.proto
```diff
@@ -99,29 +97,15 @@ message LabelKey {
...
-
-  // The data points of this timeseries. Point type MUST match the MetricDescriptor.type.
-  repeated Point points = 2;
-}
-
-// A collection of data points that describes the time-varying values
-// of a cumulative metric.
-message CumulativeTimeSeries {
-  // The time that the cumulative value was reset to zero.
-  google.protobuf.Timestamp start_time = 1; // required
```
songy23: I think we still need the start time?

bogdandrutu: Not correct, because different label-values. Am I wrong?

songy23: Quotes from our previous discussion:

> [X] Cumulative Point
> [X] should we hoist the start_time timestamp to time series?
> 1. Downside of doing so is that it's not relevant to GaugePoints. (Decision: have two TimeSeries types) (PR65)
> 2. Downside of not: timestamp is 16 bytes, the double value you're recording is half that.
> 3. Alternatives considered: Change Metric to have one of { repeated GaugeTimeSeries, or repeated CumulativeTimeSeries (with start_time) }

> [X] Should start_time be called reset_time? Or should we make a note that we are keeping consistent with Stackdriver terminology?
> 1. Decision: keep start_ and end_ despite perhaps not being perfect names.

So I think all Points will have the same start time, even with different label values, because start time here is actually the time when we start all the collecting and has nothing to do with single Point.

/cc @g-easy

bogdandrutu: I don't understand, let's assume we have 2 points with the same start_timestamp (maybe even the same timestamp).

Start timestamp: 1234sec and 56ns
Current timestamp: 1234sec 59ns
{foo, bar} -> 5.2
{foo, baz} -> 4.3

In the previous model I have to create 2 different CumulativeTimeSeries:
```
CumulativeTimeSeries: {
  start_time: {1234, 56}
  label_values: [{foo, true /* has_value */}, {bar, true}]
  points: [{timestamp: {1234, 59}, value: 5.2}]
}
CumulativeTimeSeries: {
  start_time: {1234, 56}
  label_values: [{foo, true /* has_value */}, {baz, true}]
  points: [{timestamp: {1234, 59}, value: 4.3}]
}
```

In the new proposal I have to create 2 TimeSeries objects as well:
```
TimeSeries: {
  label_values: [{foo, true /* has_value */}, {bar, true}]
  points: [{start_time: {1234, 56}, timestamp: {1234, 59}, value: 5.2}]
}
TimeSeries: {
  label_values: [{foo, true /* has_value */}, {baz, true}]
  points: [{start_time: {1234, 56}, timestamp: {1234, 59}, value: 4.3}]
}
```

As you can see I cannot save any allocation or bytes on the wire, if we want to save the bytes on the wire we should move the list of labels to be part of the Point not the TimeSeries.


What I am trying to say is that 2 is not valid in the majority of the cases.

bogdandrutu: I personally don't think that one TimeSeries is "defined" or "identify" by the start_timestamp (I know that I supported the previous decision). If we keep timestamp in the TimeSeries it means that the start_timestamp defines the TimeSeries.

songy23: Consider cases like this:

Start timestamp: 1234sec and 56ns
1st timestamp: 1234sec 59ns
{foo, bar} -> 5.2
{foo, baz} -> 4.3

2nd timestamp: 1234sec 66ns
{foo, bar} -> 5.3
{foo, baz} -> 4.4

The previous model becomes more space efficient:
```
CumulativeTimeSeries: {
  start_time: {1234, 56}
  label_values: [{foo, true /* has_value */}, {bar, true}]
  points: [
      {timestamp: {1234, 59}, value: 5.2}, 
      {timestamp: {1234, 66}, value: 5.3}
  ]
}
CumulativeTimeSeries: {
  start_time: {1234, 56}
  label_values: [{foo, true /* has_value */}, {baz, true}]
  points: [
      {timestamp: {1234, 59}, value: 4.3}, 
      {timestamp: {1234, 66}, value: 4.4}
  ]
}
```

vs.

```
TimeSeries: {
  label_values: [{foo, true /* has_value */}, {bar, true}]
  points: [
      {start_time: {1234, 56}, timestamp: {1234, 59}, value: 5.2},
      {start_time: {1234, 56}, timestamp: {1234, 66}, value: 5.3}
  ]
}
TimeSeries: {
  label_values: [{foo, true /* has_value */}, {baz, true}]
  points: [
      {start_time: {1234, 56}, timestamp: {1234, 59}, value: 4.3},
      {start_time: {1234, 56}, timestamp: {1234, 66}, value: 4.4}
  ]
}
```

As the points for a specific set of label values accumulated, the new proposal wasted more space on the duplicated start time.

> If we keep timestamp in the TimeSeries it means that the start_timestamp defines the TimeSeries.

Yes, I think cumulative TimeSeries should be defined by the start_timestamp.

g-easy: If the common case is for a `TimeSeries` to contain one `Point` (which makes sense for a periodic exporter I guess), then it doesn't matter where `start_time` lives.

Does OpenMetrics re-transmit the same `start_time` for every cumulative `Point`?

Is this `.proto` file still a conceptual data model, or do we need to start thinking about efficiency of wire encoding?

opencensus/proto/metrics/metrics.proto
```diff
@@ -135,19 +119,25 @@ message LabelValue {
 
 // A timestamped measurement.
 message Point {
-  // The moment when this point was recorded.
-  google.protobuf.Timestamp timestamp = 1; // required
+  // Available only for cumulative metrics. The time when the cumulative value
```
g-easy: > Available only for cumulative metrics

Should this be worded more strongly, like "must be present for cumulative metrics."

What happens when it's absent? "Use the previously seen value" gives us back encoding efficiency of moving `start_time` here.

opencensus/proto/metrics/metrics.proto
bogdandrutu: I need to understand what we disagree on here. If it is just where the start_time live then I can move it to TimeSeries (but this seems very strange to me) and merge the other changes first then we can argue about start_time.

opencensus/proto/metrics/metrics.proto
bogdandrutu: done.

opencensus/proto/metrics/metrics.proto
songy23: I think so - I'm fine with leaving start_time to TimeSeries and move forward with other changes.

bogdandrutu:
Merging this as we agree without moving the start_timestamp.

songy23:
LGTM

PR #73, opened 2018-08-13
Rename Distribution to Histogram.

bogdandrutu:


opencensus/proto/metrics/metrics.proto
```diff
@@ -147,13 +147,10 @@ message Point {
   }
 }
 
-// Distribution contains summary statistics for a population of values. It
-// optionally contains a histogram representing the distribution of those
-// values across a set of buckets.
-message DistributionValue {
+// Histogram contains summary statistics for a population of values.
```
songy23: In this case histogram buckets are required for `HistogramValue`?
What if users only want the summary stats (sum, count, mean, etc.) without a histogram?

songy23: SGTM

ramonza:
What's the motivation for this? Would be good to record it.

bogdandrutu:
No need for this PR because we will use our own format and the current definition is compatible with Prometheus, SD, OpenMetrics etc.

PR #74, opened 2018-08-20
Add tracestate field to the Span proto.

bogdandrutu:


opencensus/proto/trace/trace.proto
```diff
@@ -47,6 +47,23 @@ message Span {
...
+  // digits '0'-'9', underscores '_', dashes '-', asterisks '*', and forward slashes '/'. For
+  // multi-tenant vendors scenarios '@' sign can be used to prefix vendor name. The maximum length
+  // for the key is 256 characters.
+  //
+  // The value is opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the
+  // range 0x20 to 0x7E) except ',' and '='. Note that this also excludes tabs, newlines, carriage
+  // returns, etc.
+  //
+  // See the https://github.com/w3c/distributed-tracing for more details about this field.
+  map<string, string> tracestate = 15;
```
songy23: nit: consider ordering the fields by its field number. Same for `kind`.

bogdandrutu:
@lmolkova I would like you or @SergeyKanzhelev to look into this and approve it.

adriancole:
all good, just a question. are we expecting this to be used only in process
(conversion) or basically anywhere (collector tier)? I'm guessing the
latter and the size constraints would limit the impact of this, but
definitely double size of small spans. one thing possibly already addressed
is to mention only populate this on spans with remote parent (server or
message consumer)

opencensus/proto/trace/trace.proto
SergeyKanzhelev: Is `map` ordered in proto? Not that important for storing as for propagation, but nevertheless for consistency it may be better to have a list

adriancole:
another option is to switch it to a truncatable string, which retains
the order and also an help if you are trying to understand malformed
content (not entirely sure the use case)

PR #76, opened 2018-08-23
Build exporters with maven.

bogdandrutu:


songy23:
Looks like the import path doesn't work with gradle

PR #80, opened 2018-08-27
 Add trace_config.proto.

songy23:


opencensus/proto/trace/BUILD.bazel
```diff
@@ -16,7 +16,10 @@ package(default_visibility = ["//visibility:public"])
 
 proto_library(
     name = "trace_proto",
```
bogdandrutu: We should have 2 different proto_libraries here.

songy23: Got it, done.

songy23:
Thanks!

Issue #82, opened 2018-08-28
Include version number in the package path for all protos.


songy23:
Take Trace for example, the package should be "opencensus.proto.trace.v1".
Gen-go package should be "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1".

songy23:
Also removed "proto" from the gen-go package path, since the "github.com/census-instrumentation/opencensus-proto/gen-go/*" already indicates this is a proto library.

PR #83, opened 2018-08-28
Update proto and gen-go package names.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-proto/issues/82.

### DIFF
--- a/opencensus/proto/metrics/BUILD.bazel
+++ b/opencensus/proto/metrics/BUILD.bazel
@@ -16,7 +16,7 @@ package(default_visibility = ["//visibility:public"])
 
 proto_library(
     name = "metrics_proto",
-    srcs = ["metrics.proto"],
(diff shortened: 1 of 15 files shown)


opencensus/proto/metrics/BUILD.bazel
```diff
@@ -16,7 +16,7 @@ package(default_visibility = ["//visibility:public"])
 
 proto_library(
     name = "metrics_proto",
-    srcs = ["metrics.proto"],
+    srcs = ["v1/metrics.proto"],
```
bogdandrutu: Better to move the BUILD.bazel in the v1 directory otherwise we should encode the version in the bazel library name.

songy23: Make sense, done.

songy23:
Thanks!

PR #84, opened 2018-08-28
Make tracestate a list instead of a map to preserve ordering.

bogdandrutu:
One thing I want to confirm is the name tracestate vs trace_state. I don't like the fact that tracestate is not a real word so all the time the spelling checker triggers on this.

### DIFF
--- a/opencensus/proto/trace/v1/trace.proto
+++ b/opencensus/proto/trace/v1/trace.proto
@@ -47,25 +47,31 @@ message Span {
   // This field is required.
   bytes span_id = 2;
 
-  message TraceStateEntry {
-    // The key must begin with a lowercase letter, and can only contain
-    // lowercase letters 'a'-'z', digits '0'-'9', underscores '_', dashes '-',
-    // asterisks '*', and forward slashes '/'. For multi-tenant vendors
-    // scenarios '@' sign can be used to prefix vendor name. The maximum length
-    // for the key is 256 characters.
-    string key = 1;
-
-    // The value is opaque string up to 256 characters printable ASCII RFC0020
-    // characters (i.e., the range 0x20 to 0x7E) except ',' and '='. Note that
-    // this also excludes tabs, newlines, carriage returns, etc.
-    string value = 2;
-  }
-
   // This field conveys information about request position in multiple distributed tracing graphs.
-  // It is a list of TraceStateEntry with a maximum of 32 members in the list.
+  // It is a list of Tracestate.Entry with a maximum of 32 members in the list.
   //
   // See the https://github.com/w3c/distributed-tracing for more details about this field.
-  repeated TraceStateEntry trace_state_entries = 15;
+  message Tracestate {
+    message Entry {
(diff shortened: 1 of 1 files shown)


opencensus/proto/trace/v1/trace.proto
```diff
@@ -47,22 +47,25 @@ message Span {
...
+    // this also excludes tabs, newlines, carriage returns, etc.
+    string value = 2;
+  }
+
+  // This field conveys information about request position in multiple distributed tracing graphs.
+  // It is a list of TraceStateEntry with a maximum of 32 members in the list.
   //
   // See the https://github.com/w3c/distributed-tracing for more details about this field.
-  map<string, string> tracestate = 15;
+  repeated TraceStateEntry trace_state = 15;
```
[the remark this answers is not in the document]
bogdandrutu: Done.

bogdandrutu:
Blocked by one decision in https://github.com/census-instrumentation/opencensus-specs/issues/168.

opencensus/proto/trace/v1/trace.proto
```diff
@@ -47,22 +47,31 @@ message Span {
...
-  //
-  // The value is opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the
-  // range 0x20 to 0x7E) except ',' and '='. Note that this also excludes tabs, newlines, carriage
-  // returns, etc.
+  // This field conveys information about request position in multiple distributed tracing graphs.
+  // It is a list of Tracestate.Entry with a maximum of 32 members in the list.
   //
   // See the https://github.com/w3c/distributed-tracing for more details about this field.
-  map<string, string> tracestate = 15;
+  message Tracestate {
```
odeke-em: Minor nit, why don't we break this out into a message of its own, that is, a top-level definition instead of an inlined definition here?

adriancole:
yes tracestate

opencensus/proto/trace/v1/trace.proto
```diff
@@ -47,22 +47,31 @@ message Span {
...
+      // Note that this also excludes tabs, newlines, carriage returns, etc.
+      string value = 2;
+    }
+
+    // A list of entries that represent the Tracestate.
+    repeated Entry entries = 1;
+  }
+
+  // The Tracestate on the span.
+  Tracestate tracestate = 15;
```
adriancole: another option is to have a string and not parse it except to remove your incoming tracestate. OTOH what you have is more flexible for future changes. As mentioned in the spec review "For multi-tenant vendors scenarios '@' sign can be used to prefix vendor name." is highly speculative, and the format isn't well thought through. Usually '@' is not a character promoted for use for various reasons I can think of a few. I wouldn't advertise this in good faith.

adriancole: ps option for ridding `@` is just deleting the advice (ideally also from where it was copy/pasted from). We should not encourage characters that routinely require escaping.

PR #86, opened 2018-08-29
Update gen-go files.

songy23:


rakyll:
Is this breaking opencensusd? If so, can we also have a PR fixing it?

songy23:
> Is this breaking opencensusd? If so, can we also have a PR fixing it?

Yes, I will send a PR to fix the import path for OC-Service after this PR.

songy23:
Updated, PTAL.

OC-D import will be fixed with https://github.com/census-instrumentation/opencensus-service/pull/17.

songy23:
Thanks!

PR #91, opened 2018-09-04
agent/trace/v1: fix signature for Export and comments too

### LINKED ISSUE #89: agent/trace/v1: fix TraceService to articulate direction and responsibilities of agent vs applications
Currently the TraceService proto is defined as follows
https://github.com/census-instrumentation/opencensus-proto/blob/10f868359b331d9f159caf81f6ae8d1757cdaddd/opencensus/proto/agent/trace/v1/trace_service.proto#L32-L43

#### 1. The signature for Config is right but the comment is wrong

It should be that
```
// The agent pushes configs down to applications via a stream of requests
```
and not via a `stream of responses`

#### 2. The signature for Config is wrong
Config is a method that the server/service uses to send `ConfigTraceServiceRequest` values down to the applications, and the application send back `ConfigTraceServiceResponse` that the service consumes, not the other way around, thus
```proto
rpc Config(stream ConfigTraceServiceRequest) returns (stream ConfigTraceServiceResponse) {} 
```
should actually be
```proto
rpc Config(stream ConfigTraceServiceResponse) returns (stream ConfigTraceServiceRequest) {} 
```

odeke-em:
agent/trace/v1: fix signature for Config and comments too
  
* Fix the signature for Config since the service sends
down to applications on  a stream of ConfigTraceServiceRequests
and receives responses on a stream of ConfigTraceServiceResponses

* Update TraceService comments

Fixes #89

### DIFF
--- a/opencensus/proto/agent/trace/v1/trace_service.proto
+++ b/opencensus/proto/agent/trace/v1/trace_service.proto
@@ -30,14 +30,13 @@ option java_outer_classname = "TraceServiceProto";
 option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1";
 
 service TraceService {
+  // After initialization, this RPC must be kept alive for the
+  // entire life of the application. The agent pushes configs
+  // down to applications via a stream of requests.
+  rpc Config(stream ConfigTraceServiceResponse) returns (stream ConfigTraceServiceRequest) {}
 
-  // After the initialization this RPC must be kept alive for the
-  // entire life of the application. Agent pushes configs to the
-  // application via a stream of responses.
-  rpc Config(stream ConfigTraceServiceRequest) returns (stream ConfigTraceServiceResponse) {}
-
-  // Allow to export Spans.
-  // For performance reason, it is recommended to keep this RPC
+  // Allows applications to send spans to the agent.
+  // For performance reasons, it is recommended to keep this RPC
   // alive for the entire life of the application.
   rpc Export(stream ExportTraceServiceRequest) returns (stream ExportTraceServiceResponse) {}
 }


odeke-em:
@songy23 after merge, could you please help me re-generate the various gen-go/* and Java files? I ask because ideally it would be nice to have the same protoc compiler

odeke-em:
@songy23 okay, with the PR updated, you were saying we should swap their internal bodies/contents instead?
If we swapped  internal bodies, the generated signatures will look wonky still
screen shot 2018-09-04 at 2 33 04 pm

instead of
screen shot 2018-09-04 at 2 31 54 pm

where a client is sending the Config*Request instead of receiving

songy23:
I see - thanks for putting the graph, that helps me understand the issue a lot better. In this case I agree it makes more sense for the client to send *Response and recv *Request, since that's the logic of agent sending config update to library.

odeke-em:
Thanks for the review and merge @songy23! For some reason I didn't get a notification :)

opencensus/proto/agent/trace/v1/trace_service.proto
```diff
@@ -30,14 +30,13 @@ option java_outer_classname = "TraceServiceProto";
 option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1";
 
 service TraceService {
+  // After initialization, this RPC must be kept alive for the
+  // entire life of the application. The agent pushes configs
+  // down to applications via a stream of requests.
+  rpc Config(stream ConfigTraceServiceResponse) returns (stream ConfigTraceServiceRequest) {}
```
bogdandrutu: Don't understand this. Is this what we want to do? Request to be named the message that comes from the agent? I think will confuse some people but I think the idea is good. Maybe just comment that requests are coming from the agent, that's why the message sent by the agent is a "request" (or something like that to clarify).

songy23: Yes that's the idea, I posted https://github.com/census-instrumentation/opencensus-proto/pull/93 that added a note about this.

PR #92, opened 2018-09-04
Update gen-go files.

songy23:
As a follow-up of https://github.com/census-instrumentation/opencensus-proto/pull/91.


songy23:
Thanks!

PR #93, opened 2018-09-05
Add a note about config request from agent.

songy23:


opencensus/proto/agent/trace/v1/trace_service.proto
```diff
@@ -33,6 +33,8 @@ service TraceService {
   // After initialization, this RPC must be kept alive for the
   // entire life of the application. The agent pushes configs
   // down to applications via a stream of requests.
+  // Logically the requests on updating config are coming from agent,
```
odeke-em: the requests to update configurations come from the agent
and that's why the message name is suffixed with "Request"

songy23:
I picked "RequestedUpdatedConfig" as the request from Agent, and "CurrentLibraryConfig" as the response that Library sends back to Agent. Any better idea? @odeke-em @bogdandrutu

opencensus/proto/agent/trace/v1/trace_service.proto
```diff
@@ -32,16 +32,16 @@ option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/a
 service TraceService {
   // After initialization, this RPC must be kept alive for the
   // entire life of the application. The agent pushes configs
-  // down to applications via a stream of requests.
-  rpc Config(stream ConfigTraceServiceResponse) returns (stream ConfigTraceServiceRequest) {}
+  // down to applications via a stream.
+  rpc Config(stream RequestedUpdatedConfig) returns (stream CurrentLibraryConfig) {}
```
odeke-em: Nice suggestions! How about changing from 
`RequestedUpdatedConfig`
to
`LibraryConfigUpdate`

but we need to preserve the order of the streams though to
```proto
rpc Config(stream CurrentLibraryConfig) returns (stream LibraryConfigUpdate) {}
```
lest we go back to the scenario yesterday.

songy23: Oh yes, I got the order of streams wrong. The name "LibraryConfigUpdate" sounds better to me.

opencensus/proto/agent/trace/v1/trace_service.proto
```diff
@@ -32,16 +32,16 @@ option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/a
 service TraceService {
   // After initialization, this RPC must be kept alive for the
   // entire life of the application. The agent pushes configs
-  // down to applications via a stream of requests.
-  rpc Config(stream ConfigTraceServiceResponse) returns (stream ConfigTraceServiceRequest) {}
+  // down to applications via a stream.
+  rpc Config(stream CurrentLibraryConfig) returns (stream LibraryConfigUpdate) {}
```
rghetia: How about updatedLibraryConfig or newLibraryConfig instead of LibraryConfigUpdate?

songy23: Among the 3 names "UpdatedLibraryConfig" sounds the most appropriate to me, though I don't have a strong opinion. WDYT @odeke-em @bogdandrutu ?

bogdandrutu: +1 for UpdatedLibraryConfig

songy23: Thanks, looks like we have a consensus here so I renamed it to "UpdatedLibraryConfig". Also re-generated the gen-go files.

odeke-em: Nice suggestion with UpdatedLibraryConfig @rghetia, SGTM!

Issue #99, opened 2018-09-08
agent/common/v1: LibraryInfo should disambiguate between library version and exporter version


odeke-em:
On review https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/pull/7#discussion_r216053886 @bogdandrutu mentioned that `version` should actually be the library version.

The current field definition suggests that it should be version of the agent exporter. However, we need to propagate information  https://github.com/census-instrumentation/opencensus-proto/blob/ee2e6d8d42906975d77f40b1c73da8c1a58bf405/opencensus/proto/agent/common/v1/common.proto#L85-L86

### Proposal
Instead of the ambiguous `version`, let's instead make 2 fields
* oc_version -- the version of the OpenCensus implementation running
* exporter_version -- the version of the agent exporter running

bogdandrutu:
Don't know if I like the oc_version, maybe lib_version? library_version? or just version? Also because we need to be wire backwards compatible, we should keep id 2 for the exporter and add id 3 for the library.

odeke-em:
`lib*_version`, `version` are too ambiguous IMHO. As you noticed from the PR, as per the proto definitions and despite the comment
https://github.com/census-instrumentation/opencensus-proto/blob/ee2e6d8d42906975d77f40b1c73da8c1a58bf405/opencensus/proto/agent/common/v1/common.proto#L85-L86

 we interpreted things differently about `version`.

For sure, field id 2 should stay as the version of the exporter.

bogdandrutu:
maybe core_library_version? I think if we have 2 fields library_version and exporter_version will not be that confusing.

odeke-em:
Nice, `core_library_version` SGTM, thank you! Now that you mention it, having those 2 fields helps disambiguate it. However, `core_library_version` sets it in stone.

songy23:
`core_library_version ` sounds good to me too.

PR #100, opened 2018-09-08
agent/common/v1: use exporter_version, core_library_version in LibraryInfo

odeke-em:
By using:
* `exporter_version`(previously `version`, but same field id: 2)
* `core_library_version` -- version of OpenCensus Library

we can succinctly convey information to the agent about the
exporter and core libraries used.

Updates #99

### DIFF
--- a/opencensus/proto/agent/common/v1/common.proto
+++ b/opencensus/proto/agent/common/v1/common.proto
@@ -83,7 +83,10 @@ message LibraryInfo {
   Language language = 1;
 
   // Version of Agent exporter of Library.
-  string version = 2;
+  string exporter_version = 2;
+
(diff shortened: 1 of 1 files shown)


odeke-em:
/cc @songy23 @bogdandrutu.

@songy23 would you mind sending a follow-up PR to generate the `gen-go` code? I ask because for some reason using `./mkgogen.sh` from this repo can never generate working Go code for me: I get clashes about `fileDescriptor0`. Thank you.

songy23:
> @songy23 would you mind sending a follow-up PR to generate the gen-go code?

Sure, will do.

odeke-em:
Thank you for the review and merge @songy23!

PR #103, opened 2018-09-11
Re-organize proto directory structure.

songy23:
build.gradle and BUILD.bazel/WORKSPACE should not be under the same
directory, since they expect different directory layout for proto files.


bogdandrutu:
Can you re-run the gengo script and make sure nothing changes, it shouldn't but want to make sure.

songy23:
> Can you re-run the gengo script and make sure nothing changes, it shouldn't but want to make sure.

Yes, I ran `mkgogen.sh` and none of the gen-go files changed (because the proto packages didn't change, the only thing changed is the build directory layout).

PR #112, opened 2018-09-17
Add specifications on Agent implementation details.

songy23:
Updates https://github.com/census-instrumentation/opencensus-service/issues/21. 

This will be moved to https://github.com/census-instrumentation/opencensus-service/blob/master/README.md later.


src/opencensus/proto/agent/README.md
```diff
@@ -57,6 +57,58 @@ are established.
...
+proto files (for example trace_service.proto). When a library opens the connection with the
+corresponding interceptor, the first message it sends must have the `Node` identifier. The
+interceptor will then cache the `Node` for each library, and `Node` is not required for
+the subsequent messages from libraries.
+
+### Agent Core
+
+Most functionalities of Agent are in Agent Core. Agent Core's responsibilies include:
+
+1. Accept `SpanProto` from each interceptor. Note that the `SpanProto`s that are sent to
```
odeke-em: This makes it seem like we need to send `Node` for every `SpanProto`. How about
```
1. Accept SpanProto from each interceptor. Note that just like the rest of the initiating protocols, the first message that should have been sent by each interceptor MUST have Node set; Node is then cached by Agent and subsequent
SpanProto messages sent by the interceptor don't need Node. The associated Node is neccessary so that Agent core can differentiate and group SpanProto's.
```

src/opencensus/proto/agent/README.md
```diff
@@ -57,6 +57,58 @@ are established.
...
+Agent Core must have `Node` associated, so that Agent Core can differentiate and group
+`SpanProto`s by each `Node`.
+2. Store and batch `SpanProto`s.
+3. Add additional metadata/information to `SpanProto` or `Node`. For example, in a k8s
```
odeke-em: This part is confusing, what is `Node` in this instance? Whose `Node` is it? Agent Core's or the interceptor?

It is the interceptor's node and to remove the ambiguity we can perhaps update this to:
```
3. Augment each SpanProto or Node sent from the interceptor.
For example, in a Kubernetes container, Agent Core can detect the namespace, pod id
and container name and add them to its record of Node from interceptor
```

src/opencensus/proto/agent/README.md
```diff
@@ -57,6 +57,58 @@ are established.
...
+container, Agent Core can detect the namespace, pod id and container name, and add them
+to `Node`.
+4. For some configured period of time, Agent Core will push `SpanProto`s (grouped by
+`Node`s) to Exporters.
+5. Display the currently stored `SpanProto`s on local ZPages.
```
odeke-em: zPages

src/opencensus/proto/agent/README.md
songy23: IIUC the API for accepting `SpanProto` on Agent Core is stateless, so indeed the interceptor needs to send `Node` for every `SpanProto` (or batched `SpanProto`s). This all happens within the process, so we don't need to worry about the encoded size issue.

/cc @bogdandrutu

src/opencensus/proto/agent/README.md
songy23: Sounds good, updated.

src/opencensus/proto/agent/README.md
songy23: Good catch.

src/opencensus/proto/agent/README.md
odeke-em: In that case I'll defer to Bogdan to do the approval as I am only learning about the interceptors and other part right from this review.

songy23:
Thanks!

PR #119, opened 2018-09-22
Extend the TraceService service to support export/config for multiple Applications.

bogdandrutu:
This generalization of the TraceService grpc service. This change is a no-op for previously written OC library exporters or server implementations.

Any ocagent exporters does not need to change because:
  * Node is sent only in the first message because these exporters only export data for one instrumented application (Node);
  * The Node message in the UpdateTraceConfig is ignored (even if it gets received it will be the same as the application sent initially).

Any previous server implementation of this protocol does not need to change because:
  * Supposed to work directly with Applications so only one Node per RPC received.
  * The Node message in the UpdateTraceConfig is ignored by the client (if the client is an Application), so the fact that is not sent does not affect the protocol.

### DIFF
--- a/src/opencensus/proto/agent/common/v1/common.proto
+++ b/src/opencensus/proto/agent/common/v1/common.proto
@@ -27,9 +27,10 @@ option java_outer_classname = "CommonProto";
 
 option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1";
 
-// Identifier metadata of the Node that connects to OpenCensus Agent.
+// Identifier metadata of the Node (Application instrumented with OpenCensus)
+// that connects to OpenCensus Agent.
 // In the future we plan to extend the identifier proto definition to support
-// additional information (e.g cloud id, monitored resource, etc.)
+// additional information (e.g cloud id, etc.)
 message Node {
   // Identifier that uniquely identifies a process within a VM/container.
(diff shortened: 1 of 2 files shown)


src/opencensus/proto/agent/common/v1/common.proto
```diff
@@ -27,17 +27,18 @@ option java_outer_classname = "CommonProto";
...
+// Identifier metadata of the Node (Application instrumented with OpenCensus)
+// that connects to OpenCensus Agent.
 // In the future we plan to extend the identifier proto definition to support
-// additional information (e.g cloud id, monitored resource, etc.)
+// additional information (e.g cloud id, etc.)
 message Node {
   // Identifier that uniquely identifies a process within a VM/container.
   ProcessIdentifier identifier = 1;
 
   // Information on the OpenCensus Library who initiates the stream.
```
maynk: nit: who -> that?

bogdandrutu: Good catch.

bogdandrutu:
@odeke-em you are correct we want this protocol to work in case of multiplexing configs/exports.

bogdandrutu:
I am going to merge this, @SergeyKanzhelev please provide feedback on this, I know you are busy these days but I can address all the comments in a different PR.

Issue #120, opened 2018-09-25
Metrics: Finalize the metrics data model.


songy23:
Track the progress/remaining items for the metrics data model. Feel free to add more items if there are anything missing:

- [ ] Determine how the bucket boundaries should work, i.e left-closed and right-open [, ), or left-open and right closed (, ].

PR #121, opened 2018-09-27
Change histogram buckets definition to be OpenMetrics compatible.

bogdandrutu:
Fixes https://github.com/census-instrumentation/opencensus-proto/issues/120

### DIFF
--- a/src/opencensus/proto/metrics/v1/metrics.proto
+++ b/src/opencensus/proto/metrics/v1/metrics.proto
@@ -81,7 +81,8 @@ message MetricDescriptor {
     // Floating point gauge. The value can go both up and down.
     GAUGE_DOUBLE = 2;
 
-    // Distribution gauge measurement. The count can go both up and down.
+    // Distribution gauge measurement. The count and sum can go both up and
+    // down. Recorded values are always >= 0.
     // Used in scenarios like a snapshot of time the current items in a queue
     // have spent there.
     GAUGE_DISTRIBUTION = 3;
@@ -91,11 +92,12 @@ message MetricDescriptor {
     CUMULATIVE_INT64 = 4;
 
     // Floating point cumulative measurement. The value can only go up, if
-    // resets then the start_time should also be reset.
+    // resets then the start_time should also be reset. Recorded values are
+    // always >= 0.
     CUMULATIVE_DOUBLE = 5;
 
-    // Distribution cumulative measurement. The count can only go up, if resets
-    // then the start_time should also be reset.
+    // Distribution cumulative measurement. The count and sum can only go up, if
+    // resets then the start_time should also be reset.
     CUMULATIVE_DISTRIBUTION = 6;
 
     // Some frameworks implemented Histograms as a summary of observations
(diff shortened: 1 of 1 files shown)


src/opencensus/proto/metrics/v1/metrics.proto
```diff
@@ -195,26 +197,37 @@ message DistributionValue {
...
+    oneof type {
+      // Bucket with explicit bounds.
+      Explicit explict = 1;
+    }
+
+    // Specifies a set of buckets with arbitrary upper-bounds.
+    // This defines size(bounds) + 1 (= N) buckets. The boundaries for bucket
+    // index i are:
+    //
+    // [0, bucket_bounds[i]) for i == 0
```
songy23: So the histogram will only accommodate non-negative values? Is this compatible with our current APIs?

bogdandrutu: Our current APIs (in stats) do support negative values but no back-end that I investigated (Prometheus or Stackdriver) support this. So we will need to make this change in stats to only record positive values.

songy23: SGTM. In this case we need to update the gRPC metrics (https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md#units) as well.

bogdandrutu:
@songy23 not sure I understand. All values are >= 0 there.

songy23:
With this definition, the bucket bounds should not include 0:
```proto
// The values must be monotonically increasing and > 0.
repeated double bounds = 1;
```
While for the current gRPC metrics, there's a bucket (-infinity, 0] and we cannot distinguish between 0s and negative values.

bogdandrutu:
@songy23 based on the current definition of the bounds https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/stats/v1/stats.proto#L117 the first bucket is (-infinity, 0) so we just completely drop that. I think 0s are included.

src/opencensus/proto/metrics/v1/metrics.proto
```diff
@@ -195,26 +197,37 @@ message DistributionValue {
...
+    // [bucket_bounds[i-1], bucket_bounds[i]) for 0 < i < N-1
+    // [bucket_bounds[i-1], +infinity) for i == N-1
+    message Explicit {
+      // The values must be monotonically increasing and > 0.
+      repeated double bounds = 1;
+    }
+
+    // TODO: If OpenMetrics decides to support (a, b] intervals we should add
```
songy23: SGTM

src/opencensus/proto/metrics/v1/metrics.proto
odeke-em: @bogdandrutu what happens to users that would like to record temperature values and would like to OpenCensus to collect values then trigger alerts such as warming up the house? We published a blogpost such as https://opencensus.io/articles/iot/ and anyone in Canada (where I also hail from and have witnessed -40C=-40F many times and most of the year temperatures are below 0C) who might repeat that experiment might be surprised after this change. Please correct me if am wrong

bogdandrutu: @odeke-em I will give you two answers:
1. Funny answer: they can use K (Kelvin) and it is guaranteed to be greater than 0 (or if it gets to 0 it will be too late for this to matter).
2. Serious answer: currently no backend supports negative values in a distribution (Prometheus no, Stackdriver says it supports negative values, but I tested and it does not work, SignalFx not yet supports histograms). And they have good reason to not support this.

So yes that blog post can be misleading.

odeke-em: Thanks for the responses @bogdandrutu, gotcha. We'll perhaps also need to add a note in the specs repo that bucket bounds begin at 0.

g-easy:
Looking...

src/opencensus/proto/metrics/v1/metrics.proto
```diff
@@ -81,7 +81,8 @@ message MetricDescriptor {
     // Floating point gauge. The value can go both up and down.
     GAUGE_DOUBLE = 2;
 
-    // Distribution gauge measurement. The count can go both up and down.
+    // Distribution gauge measurement. The count and sum can go both up and
```
g-easy: The count can decrease?

src/opencensus/proto/metrics/v1/metrics.proto
```diff
@@ -91,11 +92,12 @@ message MetricDescriptor {
     CUMULATIVE_INT64 = 4;
 
     // Floating point cumulative measurement. The value can only go up, if
```
g-easy: The value is allowed to stay the same, right?

How about "the value cannot decrease"

src/opencensus/proto/metrics/v1/metrics.proto
```diff
@@ -195,26 +197,37 @@ message DistributionValue {
...
-  //
-  // Don't change bucket boundaries within a timeseries if your backend
-  // doesn't support this.
-  repeated double bucket_bounds = 4;
+  // If bucket_options has no type, then there is no histogram associated with
+  // the Distribution.
+  message BucketOptions {
+    oneof type {
+      // Bucket with explicit bounds.
+      Explicit explict = 1;
```
g-easy: explict -> explicit

src/opencensus/proto/metrics/v1/metrics.proto
```diff
@@ -195,26 +197,37 @@ message DistributionValue {
...
+    }
+
+    // TODO: If OpenMetrics decides to support (a, b] intervals we should add
+    // support for these by defining a boolean value here which decides what
+    // type of intervals to use.
+  }
+
+  // Don't change bucket boundaries within a TimeSeries if your backend doesn't
+  // support this. To save network bandwidth this field can be sent only the
+  // first time a metric is send when using a stream reporting request.
```
g-easy: send -> sent

g-easy: How about "stream reporting request" -> "streaming reply"

src/opencensus/proto/metrics/v1/metrics.proto
```diff
@@ -195,26 +197,37 @@ message DistributionValue {
...
+
+    // Specifies a set of buckets with arbitrary upper-bounds.
+    // This defines size(bounds) + 1 (= N) buckets. The boundaries for bucket
+    // index i are:
+    //
+    // [0, bucket_bounds[i]) for i == 0
+    // [bucket_bounds[i-1], bucket_bounds[i]) for 0 < i < N-1
+    // [bucket_bounds[i-1], +infinity) for i == N-1
+    message Explicit {
+      // The values must be monotonically increasing and > 0.
```
g-easy: + cannot repeat?

g-easy: [Wikipedia](https://en.wikipedia.org/wiki/Monotonic_function) has me worried with "a function that increases monotonically does not exclusively have to increase, it simply must not decrease"

How about "strictly increasing" ?

Issue #105, opened 2018-09-12
bazel and consumers of opencensus-proto


g-easy:
@bogdandrutu pointed me at https://github.com/census-instrumentation/opencensus-proto/pull/103 which adds a `src/` dir.

This issue is to investigate what this looks like for projects that use bazel and depend on `opencensus-proto`, e.g. do `deps` now look like `"@opencensus_proto//src/opencensus/proto/trace/v1:trace_proto"`

songy23:
I believe #103 shouldn't change the Bazel deps for other projects, because the `WORKSPACE` file also moves to `src/`: https://github.com/census-instrumentation/opencensus-proto/blob/d3b7200dbd14743faa22c6642291d76030833cbd/src/WORKSPACE#L15

Now `src` becomes the root directory and the deps should still be  "@opencensus_proto//opencensus/proto/trace/v1:trace_proto".

g-easy:
Two alternatives:
```
http_archive(
    name = "io_opencensus_proto",
    strip_prefix = "opencensus-proto-master",
    urls = ["https://github.com/census-instrumentation/opencensus-proto/archive/master.zip"],
)

http_archive(
    name = "io_opencensus_proto_2",
    strip_prefix = "opencensus-proto-master/src",
    urls = ["https://github.com/census-instrumentation/opencensus-proto/archive/master.zip"],
)
```

Both of these work:
```shell
bazel build @io_opencensus_proto//src/opencensus/proto/trace/v1:trace_proto
bazel build @io_opencensus_proto_2//opencensus/proto/trace/v1:trace_proto
```

I'm currently trying to build an example that imports trace.proto

g-easy:
It doesn't look like bazel is paying attention to the WORKSPACE file in opencensus-proto:

In the first case, the `src/` subdir has to be in the path.

In both cases it doesn't warn (but I'm sure I've seen it do this in the past!) that my archive name (io_opencensus_etc) doesn't match the subrepository's name (opencensus_proto)

g-easy:
Note that in `http_archive()` io_opencensus_proto_2, I strip the `/src` component, but `git_repository()` doesn't have a strip_prefix argument.

g-easy:
Here's a working `cc_grpc_library()` example that consumes opencensus-protos: https://github.com/g-easy/oc-proto-example

PR #140, opened 2018-10-12
README: Add instructions on using opencensus_proto with Bazel.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-proto/issues/105.


songy23:
> Maybe an example with git_archive is good to have as well

SG, added.

PR #141, opened 2018-10-22
Web interaction proto

draffensperger:
This is a proto designed specifically for web clients to send user interaction measurements to the an OpenCensus agent interceptor (see https://github.com/census-instrumentation/opencensus-service).

This format is similar to the main Span proto but is designed to make it easy for web clients to send data in a form friendly for JSON representation and to allow it to send raw browser performance entries from `window.performance.getEntries()`


draffensperger:
cc/ @fabxc @rakyll @jkohen

PR #145, opened 2018-10-30
[WIP] Add error proto definitions

lmolkova:
This change adds error signal to opencensus: https://github.com/census-instrumentation/opencensus-specs/issues/154

TODO:

- [ ] Agree on raw and parsed stack trace design:

* Having raw stack trace is beneficial for all backends that support exceptions.
* Parsed stack is needed for advanced scenarios only. E.g. Jaeger likely does not need parsed one, and StackDriver has parser on the backend
Going forward, having stack parsing implemented in one place (agent exporter) and for specific language (i.e using helper tools that exists in this language) is more sustainable approach then making all backends parse all language-specific errors.
On the other side, backends that do not need parsed stack would still need raw one and making them format string from it is not sustainable again.
**Proposal**: by default send both, make it configurable by agent/service

- [ ] Java suppressed exceptions vs cause - there is no place for suppressed exceptions. do we want to have it now or wait until somebody requests them?

- [ ]  common code: `StackTrace`, `AttributeValue`, `TruncatableString` and `Module` are common between Spans and Errors. Does it make sense to have another common proto for them?

Here is how exceptions/errors are mapped to the proto definitions in a few languages:

# Java Exception
| Throwable property name | Proto property | Language-specific description | 
|------------------------|----------------|-------------|
| message | message | message that describes the current exception | 
| cause | causes | linked list of instances that caused the current exception |
| stackTrace | parsed_stack_trace.stack_frames & raw_stack_trace  | immediate frames on the call stack |
| suppressed | ??? TODO | exceptions that were suppressed in order to deliver this exception |
| throwable.getClass() | error_info.error_type | type of Throwable |

# C# Exception
| Exception property name | Proto property | Language-specific description | 
|------------------------|----------------|-------------|
| Data | attributes | collection of additional user-defined key/value pairs | 
| HelpLink | attributes | link to the help file associated with this exception | 
| HResult | attributes | a coded numerical value that is assigned to a specific exception | 
| InnerException | causes | linked list of instances that caused the current exception |
| Message | message | message that describes the current exception | 
| Source | parsed_stack_trace.stack_frames.frame[0].load_module | application or the object that causes the error |
| StackTrace | parsed_stack_trace (parsed from string) & raw_stack_trace  | string representation of the immediate frames on the call stack |
| TargetSite |  already in the stack trace | method handle that throws the current exception |
| typeof(Exception) | error_info.error_type | type of Exception |

# Python exception
| Property name | Proto property | Language-specific description | 
|------------------------|----------------|-------------|
| exception.message, sys.exc_info() | message | message that describes the current exception | 
| traceback.format_exc(). | parsed_stack_trace.stack_frames (parsed from string []) & raw_stack_trace  | formatted stack trace | 
| type(exception), sys.exc_info() | error_info.error_type | type of exception |

# Go error
| Property name | Proto property | Language-specific description | 
|------------------------|----------------|-------------|
| message | message | message that describes the current exception | 
| runtime.Stack() | parsed_stack_trace.stack_frames (parsed from byte[]) & raw_stack_trace | formatted stack trace of the goroutine | 
| err.(type) |  error_info.error_type | type of Error |

@bogdandrutu  @adriancole @songy23 @rakyll @odeke-em @SergeyKanzhelev please review


lmolkova:
ping @bogdandrutu @adriancole @songy23 @rakyll @odeke-em @SergeyKanzhelev

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+  google.protobuf.Timestamp time = 1;
+
+  // Error description such as exception message.
+  // This field is required.  
+  string message = 2;
+
+  // TraceId of the span in which scope the error occurred. 
+  // Must be set if there is a current span or it was explicitly
+  // provided by the caller
+  bytes trace_id = 3;
```
bogdandrutu: Should we add a SpanContext message? Also should we include the Tracestate?

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+    //     "custom.attribute": ""
+    map<string, AttributeValue> attribute_map = 1;
+
+    // The number of attributes that were discarded. Attributes can be discarded
+    // because their keys are too long or because there are too many attributes.
+    // If this value is 0, then no attributes were dropped.
+    int32 dropped_attributes_count = 2;
+  }
+  
+  // A list of attributes on the error
```
bogdandrutu: Is this needed? How are these fields used?

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+    int64 int_value = 2;
+    // A Boolean value represented by `true` or `false`.
+    bool bool_value = 3;
+    // A double value.
+    double double_value = 4;
+  }
+}
+
+// The call stack which originated this span.
+message StackTrace {
```
bogdandrutu: This can be shared between trace and error.

bogdandrutu:
@SergeyKanzhelev @lmolkova I think I made a mistake about naming. Should we go with singular or plural for package names (metric**s** vs error). Let's make a decision now and try to be consistent if possible.

src/opencensus/proto/error/v1/error.proto
SergeyKanzhelev: `SpanContext` sounds reasonable. And it will be good to have. One issue with this is inconsistency with spans. In span - there is a `parent_span_id` defined as a separate field. Also for logs the current thinking is separate fields (my understanding): https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/LogCorrelation.md#log-correlation

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+option java_package = "io.opencensus.proto.error.v1";
+option java_outer_classname = "ErrorProto";
+
+option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/error/v1";
+
+// Error event that is generated by integration or application.
+message Error {
+  // Time when the error occurred.
+  // This field is required.  
+  google.protobuf.Timestamp time = 1;
```
SergeyKanzhelev: comment that this field is required

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+
+  // Error description such as exception message.
+  // This field is required.  
+  string message = 2;
```
SergeyKanzhelev: Is message required?

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+
+  // TraceId of the span in which scope the error occurred. 
+  // Must be set if there is a current span or it was explicitly
+  // provided by the caller
+  bytes trace_id = 3;
+
+  // SpanId of the span in which scope the error occurred. 
+  // Must be set if there is a current span or it was explicitly
+  // provided by the caller
+  bytes span_id = 4;
```
SergeyKanzhelev: should it be `parent_span_id` to align with the span definition?

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+
+    // Describes stack trace in the format convenient for specific language and exporter
+    oneof raw_stack_trace
+    {
+      // Raw string value returned by: 
+      // Java: Must be the return value of Throwable.printStackTrace().
+      // Python: Must be the return value of traceback.format_exc().
+      // JavaScript: Must be the value of error.stack as returned by V8.
+      // Ruby: Must contain joined frames returned by Exception.backtrace.
+      // C#: Must be the return value of Exception.ToString().
```
SergeyKanzhelev: Perhaps `new StackTrace(Exception).ToString()`? ToString will include message.

Also, in C# there are two stacks - unwind stack which exception was thrown thru and current thread stack. Perhaps those should be merged

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  // A collection of stack frames, which can be truncated.
+  message StackFrames {
+    // Stack frames in this call stack.
+    repeated StackFrame frame = 1;
+    // The number of stack frames that were dropped because there
+    // were too many stack frames.
+    // If this value is 0, then no stack frames were dropped.
+    int32 dropped_frames_count = 2;
```
SergeyKanzhelev: dropped from the top or from the bottom? Perhaps there should be two numbers

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+  }
+}
+
+// The call stack which originated this span.
+message StackTrace {
+  // A single stack frame in a stack trace.
+  message StackFrame {
+    // The fully-qualified name that uniquely identifies the function or
+    // method that is active in this frame.
+    TruncatableString function_name = 1;
```
SergeyKanzhelev: this should be marked as required

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+
+option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/error/v1";
+
+// Error event that is generated by integration or application.
+message Error {
+  // Time when the error occurred.
+  // This field is required.  
+  google.protobuf.Timestamp time = 1;
+
+  // Error description such as exception message.
```
SergeyKanzhelev: Should this say that message must be in English where possible?

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+  // Subsequent spans within the same request can refer
+  // to that stack trace by setting only `stack_trace_hash_id`.
+  //
+  // TODO: describe how to deal with the case where stack_trace_hash_id is
+  // zero because it was not set.
+  uint64 stack_trace_hash_id = 2;
+}
+
+// A description of a binary module.
+message Module {
```
SergeyKanzhelev: perhaps add Attributes to the module. So things like `MVID`, `filepath`, `file version` can be expressed which may be language specific

src/opencensus/proto/error/v1/error.proto
SergeyKanzhelev: I believe this file is a copy looking at content

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+  // Often multiple spans will have identical stack traces.
+  // The first occurrence of a stack trace should contain both
+  // `stack_frames` and a value in `stack_trace_hash_id`.
+  //
+  // Subsequent spans within the same request can refer
+  // to that stack trace by setting only `stack_trace_hash_id`.
+  //
+  // TODO: describe how to deal with the case where stack_trace_hash_id is
+  // zero because it was not set.
+  uint64 stack_trace_hash_id = 2;
```
SergeyKanzhelev: should stackFrames be de-dupe-able

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+    TruncatableString file_name = 3;
+    // The line number in `file_name` where the function call appears.
+    int64 line_number = 4;
+    // The column number where the function call appears, if available.
+    // This is important in JavaScript because of its anonymous functions.
+    int64 column_number = 5;
+    // The binary module from where the code was loaded.
+    Module load_module = 6;
+    // The version of the deployed source code.
+    TruncatableString source_version = 7;
```
SergeyKanzhelev: is it a commit version?

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+  // List of exceptions that cause the error.
+  CauseErrors causes = 6;
+  
+  // Error severity.
+  ErrorSeverity severity = 7;
+
+  //TODO: can we have common code with trace proto?
+  // A set of attributes, each with a key and a value.
+  message Attributes {
+    // The set of attributes. The value can be a string, an integer, or the
```
SergeyKanzhelev: latest trace spec allows `double` as well

src/opencensus/proto/error/v1/error.proto
```diff
@@ -0,0 +1,245 @@
...
+    bool bool_value = 3;
+    // A double value.
+    double double_value = 4;
+  }
+}
+
+// The call stack which originated this span.
+message StackTrace {
+  // A single stack frame in a stack trace.
+  message StackFrame {
```
SergeyKanzhelev: just a random thought - if argument values were available - should those be included?

SergeyKanzhelev: Including ordering number may be helpful. If not - at least comment on how stack trace is ordered (lowest frame first or last?).

punya:
Closing as stale/obsolete.

PR #153, opened 2018-12-04
Add git_repository import to fix deprecation warning

draffensperger:
This is an attempt to fix the failing build for https://github.com/census-instrumentation/opencensus-proto/pull/77

### DIFF
--- a/src/WORKSPACE
+++ b/src/WORKSPACE
@@ -15,6 +15,7 @@
 workspace(name = "opencensus_proto")
 
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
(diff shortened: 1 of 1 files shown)


src/WORKSPACE
```diff
@@ -14,6 +14,8 @@
 
 workspace(name = "opencensus_proto")
 
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
```
songy23: Add http_archive also:
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

draffensperger:
It turns out that even after adding the `http_archive` import the bazel build still failed. An updated suggestion for how to fix the build is at https://github.com/census-instrumentation/opencensus-proto/pull/154

PR #154, opened 2018-12-04
Fix bazel build after bazel upgrade

draffensperger:
The bazel build was failing due to a variety of deprecation issues. 

This imports `git_repository` and `http_archive` from `@bazel_tools//tools/build_defs/repo:http.bzl`. This also required an upgrade to `@io_bazel_rules_go`, which then breaks `@org_pubref_rules_protobuf//go:rules.bzl`. 

To fix that, this switches to using `go_proto_library` from `@io_bazel_rules_go//proto:def.bzl`, which required some changes to the build params. This also combines the trace and trace config protos into a unified `trace_proto` build, which fixed a compile error for `//opencensus/proto/agent/trace/v1:trace_service_proto_go`.

### DIFF
--- a/src/WORKSPACE
+++ b/src/WORKSPACE
@@ -62,32 +62,9 @@ http_archive(
 git_repository(
     name = "io_bazel_rules_go",
     remote = "https://github.com/bazelbuild/rules_go.git",
-    # we need rules_go of tag 0.12.1
-    # which includes golang protobuf that supports "paths=source_relative"
-    # the "paths=source_relative" option for protoc-gen-go makes bazel works as expect
-    # when go_package was declared in proto files
-    # see https://github.com/bazelbuild/rules_go/blob/release-0.12/go/private/repositories.bzl#L75
-    # for the included golang protobuf version and
-    # see https://github.com/golang/protobuf/pull/544 for "paths=source_relative" usage
-    tag = "0.12.1",
+    tag = "0.15.8",
 )
 
 load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies", "go_register_toolchains")
 go_rules_dependencies()
 go_register_toolchains()
-
-# golang protobuf rules related
-git_repository(
-    name = "org_pubref_rules_protobuf",
-    commit = "5cae42382b620aa1e347ecf30b3e92fd0d97998c", # Jun 23, 2018
-    remote = "https://github.com/pubref/rules_protobuf.git",
-)
-
-load("@org_pubref_rules_protobuf//go:rules.bzl", "go_proto_repositories")
-go_proto_repositories(
-  excludes = [
-        # we have specified protobuf above
-        # so ignore it in case version collision
-        "com_google_protobuf",
-  ],
-)
--- a/src/WORKSPACE
+++ b/src/WORKSPACE
@@ -15,6 +15,7 @@
 workspace(name = "opencensus_proto")
 
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 # Import gRPC git repo so that we can load java_grpc_library build.
 git_repository(
--- a/src/WORKSPACE
+++ b/src/WORKSPACE
@@ -62,6 +62,13 @@ http_archive(
 git_repository(
     name = "io_bazel_rules_go",
     remote = "https://github.com/bazelbuild/rules_go.git",
+    # we need rules_go of tag 0.12.0 or greater
+    # which includes golang protobuf that supports "paths=source_relative"
+    # the "paths=source_relative" option for protoc-gen-go makes bazel works as expect
+    # when go_package was declared in proto files
+    # see https://github.com/bazelbuild/rules_go/blob/release-0.12/go/private/repositories.bzl#L75
+    # for the included golang protobuf version and
+    # see https://github.com/golang/protobuf/pull/544 for "paths=source_relative" usage
     tag = "0.15.8",
 )
 
--- a/src/WORKSPACE
+++ b/src/WORKSPACE
@@ -14,6 +14,8 @@
 
 workspace(name = "opencensus_proto")
 
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
 # Import gRPC git repo so that we can load java_grpc_library build.
 git_repository(
     name = "grpc_java",
--- a/src/WORKSPACE
+++ b/src/WORKSPACE
@@ -62,14 +62,14 @@ http_archive(
 git_repository(
     name = "io_bazel_rules_go",
     remote = "https://github.com/bazelbuild/rules_go.git",
-    # we need rules_go of tag 0.12.0
+    # we need rules_go of tag 0.12.1
     # which includes golang protobuf that supports "paths=source_relative"
     # the "paths=source_relative" option for protoc-gen-go makes bazel works as expect
     # when go_package was declared in proto files
     # see https://github.com/bazelbuild/rules_go/blob/release-0.12/go/private/repositories.bzl#L75
     # for the included golang protobuf version and
     # see https://github.com/golang/protobuf/pull/544 for "paths=source_relative" usage
-    tag = "0.12.0",
+    tag = "0.12.1",
 )
 
 load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies", "go_register_toolchains")
--- a/src/opencensus/proto/resource/v1/BUILD.bazel
+++ b/src/opencensus/proto/resource/v1/BUILD.bazel
@@ -14,7 +14,7 @@
 
 package(default_visibility = ["//visibility:public"])
 
-load("@org_pubref_rules_protobuf//go:rules.bzl", "go_proto_library")
+load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")
 
 proto_library(
     name = "resource_proto",
@@ -33,9 +33,6 @@ java_proto_library(
 
 go_proto_library(
     name = "resource_proto_go",
-    protos = ["resource.proto"],
-    pb_options = [
-       # omit the go_package declared in proto files to make bazel works as expect
-       "paths=source_relative",
-    ],
+    proto = ":resource_proto",
+    importpath = "github.com/census-instrumentation/opencensus-proto/gen-go/resource/v1",
 )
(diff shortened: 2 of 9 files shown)


src/WORKSPACE
```diff
@@ -59,32 +62,9 @@ http_archive(
 git_repository(
     name = "io_bazel_rules_go",
     remote = "https://github.com/bazelbuild/rules_go.git",
-    # we need rules_go of tag 0.12.0
```
bogdandrutu: Please keep the comment about usage of source_relative.

src/opencensus/proto/resource/v1/BUILD.bazel
```diff
@@ -33,9 +33,6 @@ java_proto_library(
 
 go_proto_library(
     name = "resource_proto_go",
-    protos = ["resource.proto"],
-    pb_options = [
-       # omit the go_package declared in proto files to make bazel works as expect
```
bogdandrutu: why did you remove this?

src/opencensus/proto/trace/v1/BUILD.bazel
```diff
@@ -14,66 +14,36 @@
...
+        "trace.proto",
+        "trace_config.proto"
+    ],
     deps = [
         "@com_google_protobuf//:timestamp_proto",
         "@com_google_protobuf//:wrappers_proto",
     ],
 )
 
-proto_library(
```
bogdandrutu: Why do you do this? It should work with having them in different entries.

src/WORKSPACE
draffensperger: Done.

src/opencensus/proto/trace/v1/BUILD.bazel
draffensperger: When the trace and trace_config rules are separate there is an error building the `//opencensus/proto/agent/trace/v1:trace_service_proto_go` rule. See: https://travis-ci.org/census-instrumentation/opencensus-proto/builds/463521132?utm_source=github_status&utm_medium=notification

I suspect it's because there are two `go_proto_library` rules for `trace_proto_go` and `trace_config_proto_go`, but they both share the same `importpath`. Combining them into a single rule allows the build to pass.

Am I missing something here? Any thoughts on how to do this better?

draffensperger: I think I need to create a separate Go proto build rule with the `paths=source_relative` flag and then pass it to the `compiler` option of `go_proto_library`.

I'll plan to work on that some time soon.

src/opencensus/proto/resource/v1/BUILD.bazel
draffensperger: I think I need to create a separate Go proto build rule with the paths=source_relative flag and then pass it to the compiler option of go_proto_library.

I'll plan to work on that some time soon.

src/opencensus/proto/trace/v1/BUILD.bazel
draffensperger: Slight correction, the relevant failing Travis build is https://travis-ci.org/census-instrumentation/opencensus-proto/jobs/463529698

It says `undefined: "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1".TraceConfig`

songy23: I think it's better to keep `trace.proto` and `trace_config.proto` separate since in most cases users only need `trace.proto`.

cc @yancl for suggestions on go_proto_library.

yancl: I thinks maybe this is a bug of **io_bazel_rules_go**(which filtered out the trace_config.proto).

I made the following diagnose:
* `bazel build -s //...` to output the sub commands with the **-s** options
* it outputs something like this:
```
(cd /home/vagrant/.cache/bazel/_bazel_vagrant/39f7101e26fcabfa22265379537b720c/execroot/opencensus_proto && \
  exec env - \
    CGO_ENABLED=1 \
    GOARCH=amd64 \
    GOOS=linux \
    GOROOT=external/go_sdk \
    GOROOT_FINAL=GOROOT \
    PATH=/usr/bin:/bin \
  bazel-out/host/bin/external/io_bazel_rules_go/go/tools/builders/linux_amd64_stripped/compile **-v** -sdk external/go_sdk ...
```
* I add a **-v** args to the above command to output the real `go tool compile` commands like the following:
```
CGO_ENABLED=1 \
GOARCH=amd64 \
GOOS=linux \
GOROOT=external/go_sdk \
GOROOT_FINAL=GOROOT \
PATH=/usr/bin:/bin \
external/go_sdk/pkg/tool/linux_amd64/compile -importcfg /home/vagrant/.cache/bazel/_bazel_vagrant/39f7101e26fcabfa22265379537b720c/execroot/opencensus_proto/bazel-out/k8-fastbuild/bin/opencensus/proto/agent/trace/v1/linux_amd64_stripped/trace_service_proto_go%/github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/importcfg436570685 -pack -o /home/vagrant/.cache/bazel/_bazel_vagrant/39f7101e26fcabfa22265379537b720c/execroot/opencensus_proto/bazel-out/k8-fastbuild/bin/opencensus/proto/agent/trace/v1/linux_amd64_stripped/trace_service_proto_go%/github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1.a -trimpath /home/vagrant/.cache/bazel/_bazel_vagrant/39f7101e26fcabfa22265379537b720c/execroot/opencensus_proto -p github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1 -- /home/vagrant/.cache/bazel/_bazel_vagrant/39f7101e26fcabfa22265379537b720c/execroot/opencensus_proto/bazel-out/k8-fastbuild/bin/opencensus/proto/agent/trace/v1/linux_amd64_stripped/trace_service_proto_go%/github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1/trace_service.pb.go
```
* As you see there is **-importcfg** args, then I find the **trace/v1.a** in the depends file:
```
$ cat /home/vagrant/.cache/bazel/_bazel_vagrant/39f7101e26fcabfa22265379537b720c/execroot/opencensus_proto/bazel-out/k8-fastbuild/bin/opencensus/proto/agent/trace/v1/linux_amd64_stripped/trace_service_proto_go%/github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/importcfg722808585
packagefile fmt=/home/vagrant/.cache/bazel/_bazel_vagrant/39f7101e26fcabfa22265379537b720c/execroot/opencensus_proto/external/go_sdk/pkg/linux_amd64/fmt.a
packagefile math=/home/vagrant/.cache/bazel/_bazel_vagrant/39f7101e26fcabfa22265379537b720c/execroot/opencensus_proto/external/go_sdk/pkg/linux_amd64/math.a
...

packagefile github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1=/home/vagrant/.cache/bazel/_bazel_vagrant/39f7101e26fcabfa22265379537b720c/execroot/opencensus_proto/bazel-out/k8-fastbuild/bin/opencensus/proto/trace/v1/linux_amd64_stripped/trace_proto_go%/github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1.a
...
```

* Then I dump the symbols of the **v1.a** using golang **nm** and can't find **TraceConfig**
```
$ ./bazel-src/external/go_sdk/pkg/tool/linux_amd64/nm /home/vagrant/.cache/bazel/_bazel_vagrant/39f7101e26fcabfa22265379537b720c/execroot/opencensus_proto/bazel-out/k8-fastbuild/bin/opencensus/proto/trace/v1/linux_amd64_stripped/trace_proto_go%/github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1.a | grep TraceConfig
vagrant@vbox:~/3rd/opencensus-proto/src$
```

* But with the correct **v1.a**  build by @draffensperger previous branch(fix-bazel-build2) I can find the **TraceConfig**
```
$ ./bazel-src/external/go_sdk/pkg/tool/linux_amd64/nm ~/3rd/x/opencensus-proto/src/bazel-bin/opencensus/proto/trace/v1/linux_amd64_stripped/trace_proto_go%/github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1.a | grep TraceConfig | head -n 10
   35a87 T %22%22.(*TraceConfig).Descriptor
   35faf T %22%22.(*TraceConfig).GetConstantSampler
```

So maybe I need some more debug for why the trace_config.proto filtered out if I was right:)

draffensperger: Thanks for the investigation here. I posted an issue for this as well: https://github.com/bazelbuild/rules_go/issues/1841

One workaround would be to create a new separate rule `proto_library` called `trace_and_config` that combines the trace.proto and trace_config.proto, and then depend on that one for the trace agent service. We could comment to explain this issue.

src/opencensus/proto/resource/v1/BUILD.bazel
draffensperger: It seems like relative import paths are not yet supported by the `go_proto_library` from io_bazel_rules_go. See https://github.com/bazelbuild/rules_go/issues/1645

Why did we choose relative import paths in the first place?

src/WORKSPACE
draffensperger: It turns out that the `paths=source_relative` option does not work with the `go_proto_library` provided by io_bazel_rules_go. See: https://github.com/bazelbuild/rules_go/issues/1842

Would it be OK to not use source_relative and so remove this comment? Are there any dependencies of `opencensus-proto` we need to make sure build correctly without it?

src/opencensus/proto/trace/v1/BUILD.bazel
draffensperger: Based on feedback on that issue (https://github.com/bazelbuild/rules_go/issues/1841#issuecomment-444665833), it seems like the best approach is to use an intermediate `go_library` rule that embeds the trace and trace config `go_proto_library`'s, and then depend on that in the trace agent service. I did that and added a comment with a link to the GitHub issue.

src/WORKSPACE
songy23: /cc @yancl Does go_proto_library with absolute path work with your build? If so then LGTM.

src/opencensus/proto/trace/v1/BUILD.bazel
draffensperger: And thanks @yancl for the detailed investigation!

src/WORKSPACE
yancl: LGTM:)

PR #160, opened 2019-01-04
Minor comments fixes

SergeyKanzhelev:


src/opencensus/proto/trace/v1/trace.proto
```diff
@@ -26,7 +26,7 @@ option java_outer_classname = "TraceProto";
 option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1";
 
 // A span represents a single operation within a trace. Spans can be
-// nested to form a trace tree. Often, a trace contains a root span
+// nested to form a trace graph. Often, a trace contains a root span
```
odeke-em: Please revert this, "graph" is the general case and allows cycles i.e. it would imply that span A could be the parent of span B and span B could the parent of span A; tree is a directed acyclic graph so the wording was correct before, unless I missed something /cc @bogdandrutu

src/opencensus/proto/trace/v1/trace.proto
```diff
@@ -36,13 +36,15 @@ option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/t
 // TODO(bdrutu): Add an example.
 message Span {
   // A unique identifier for a trace. All spans from the same trace share
-  // the same `trace_id`. The ID is a 16-byte array.
+  // the same `trace_id`. The ID is a 16-byte array. All zeroes ID is
```
odeke-em: "An ID with all zeroes is considered incorrect"

src/opencensus/proto/trace/v1/trace.proto
```diff
@@ -36,13 +36,15 @@ option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/t
...
-  // the same `trace_id`. The ID is a 16-byte array.
+  // the same `trace_id`. The ID is a 16-byte array. All zeroes ID is 
+  // considered incorrect.
   //
   // This field is required.
   bytes trace_id = 1;
 
   // A unique identifier for a span within a trace, assigned when the span
-  // is created. The ID is an 8-byte array.
+  // is created. The ID is an 8-byte array. All zeroes ID is considered
```
odeke-em: Ditto about "An ID with all zeroes"

src/opencensus/proto/trace/v1/trace.proto
```diff
@@ -217,12 +220,11 @@ message Span {
   // where a single batch handler processes multiple requests from different
   // traces or when the handler receives a request from a different project.
   message Link {
-    // A unique identifier for a trace. All spans from the same trace share
-    // the same `trace_id`. The ID is a 16-byte array.
+    // A unique identifier of a trace this linked span is part of. The ID is a
```
odeke-em: "that this linked span is a part of."

src/opencensus/proto/trace/v1/trace.proto
SergeyKanzhelev: `Links` make trace a graph, not a tree. Without `Links` it will be a tree

odeke-em: Right, but currently the comment just mentions spans and not links, of which your change perhaps assumes that the user knows about links and I am not sure that this was meant to introduce links so far.

PR #172, opened 2019-01-16
Add tags package based on the specs.

bogdandrutu:


src/opencensus/proto/tags/v1/tags.proto
```diff
@@ -0,0 +1,79 @@
...
+
+option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/tags/v1";
+
+option java_multiple_files = true;
+option java_package = "io.opencensus.proto.tags.v1";
+option java_outer_classname = "TagsProto";
+
+// TagScope is used to determine the scope of a Tag.
+message TagScope {
+  enum Scope {
```
SergeyKanzhelev: Is this information useful for exporter?

src/opencensus/proto/tags/v1/tags.proto
```diff
@@ -0,0 +1,79 @@
...
+  // This field is optional, if not present the default scope is REQUEST_SCOPE
+  // (the motivation is backwards compatibility).
+  TagScope scope = 3;
+}
+
+// TagMap is an abstract data type that represents collection of tags. i.e.,
+// each key is associated with exactly one value.
+message TagMap {
+  // Not a map becuase the proto3 map does not allow a message as the key.
+  repeated Tag tags = 1;
```
SergeyKanzhelev: so tags with the same name is allowed, correct? Perhaps something should be said about it in comment

src/opencensus/proto/tags/v1/tags.proto
```diff
@@ -0,0 +1,79 @@
...
+  // This field is required.
+  string value = 1;
+}
+
+// TagValue is the value associated with a Tag.
+message TagValue {
+  // It MUST contain only printable ASCII (codes between 32 and 126)
+  //
+  // This field is required.
+  string value = 1;
```
SergeyKanzhelev: why not truncatable string? If it is not a truncatable string - what's the reason to have a message for it and not simply use `string` as before? Same comment for `TagKey`

src/opencensus/proto/tags/v1/tags.proto
SergeyKanzhelev: My understanding - this information is only needed for API to decide what to do with the tag

src/opencensus/proto/tags/v1/tags.proto
bogdandrutu: The list should contains unique TagKeys. The proto3 map does not allow the map_key to be a proto message see https://developers.google.com/protocol-buffers/docs/proto3#maps.

src/opencensus/proto/tags/v1/tags.proto
bogdandrutu: The reason to have a message:
1. For the tag key in the future we may add other things - e.g. type information (string, int, etc.) or any other metadata.
2. For the tag value in the future we may support other types like int.

src/opencensus/proto/tags/v1/tags.proto
bogdandrutu: Correct, this is not needed for the exporter, but I defined it here for few reasons:
1. To have the language independent definition of the data model.
2. In case anyone wants to use a static config (json file to define tags that are set).

bogdandrutu:
@SergeyKanzhelev also see this https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagMap.md

src/opencensus/proto/tags/v1/tags.proto
SergeyKanzhelev: Do you want exporters to set the `scope` value? If it is no use in agent and collector - there should be a comment saying that this field is only needed for data generation and tags defining cases, not for exporting?

src/opencensus/proto/tags/v1/tags.proto
SergeyKanzhelev: The fact that the same key value is not allowed is not clear from the comment at all. As minimum I suggest to indicate this requirement in a comment.

SergeyKanzhelev: BTW, `map` with the `string` key is used for `Attributes` already. Arguably `Attributes` keys should have all the same scenarios as tags.

SergeyKanzhelev: Just to clarify - some misunderstanding may be coming from how you can read the sentence "each key is associated with exactly one value.". If you treat "each key" as unique string - it is already handled. But I read it as one key object maps to one value object. I see you already handled it, just clarifying what I meant,

src/opencensus/proto/tags/v1/tags.proto
bogdandrutu: done.

src/opencensus/proto/tags/v1/tags.proto
bogdandrutu: The reason of using a TagKey instead of directly string is because few reasons:
1) Later we can make the TagKey { oneof (string, int) } where int may represent a global id if you have a global registry for all the tags.
2) For the string in TagKey we need to do validation (limited char set) so this way people can create the key once and avoid validation every time in the api.

src/opencensus/proto/tags/v1/tags.proto
```diff
@@ -0,0 +1,82 @@
...
+
+// TagKey is the name of the Tag.
+message TagKey {
+  // Restrictions:
+  //   * MUST contain only printable ASCII (codes between 32 and 126 inclusive)
+  //   * MUST have length greater than zero and less than 256.
+  //   * MUST not be empty.
+  //
+  // This field is required.
+  string value = 1;
```
mayurkale22: I think this should be like -> ```string name = 1;```

src/opencensus/proto/tags/v1/tags.proto
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// TagValue is the value associated with a Tag.
+message TagValue {
+  // It MUST contain only printable ASCII (codes between 32 and 126)
+  //
```
mayurkale22: It cannot be longer than 256 right? https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/tags/TagValue.java#L49

Issue #169, opened 2019-01-15
Opencensus proto Spans should optionally contain Node/Process information.


sjkaris:
### Problem
Opencensus-proto [Spans](https://github.com/census-instrumentation/opencensus-proto/blob/master/gen-go/trace/v1/trace.pb.go#L135) are the building block for transmitting oc span data. With the current data models, there is no way to transmit Node information without using the agent's [ExportTraceServiceRequest](https://github.com/census-instrumentation/opencensus-proto/blob/master/gen-go/agent/trace/v1/trace_service.pb.go#L130) which disallows batching of spans from different nodes. This batching strategy is suboptimal for collector or agent processes that are accepting spans from many nodes. This additionally hampers potential benefits from sending compressed batches of spans, since it limits the batches to that of spans from a single node.

### Proposed Solution
In order to allow for arbitrary batching of Spans, I propose adding optional Node/Process information as a field on the span proto structure. To do this the following changes would be needed:

- [Node](https://github.com/census-instrumentation/opencensus-proto/blob/master/gen-go/agent/common/v1/common.pb.go#L77) is promoted up out of the agent proto definition, and into the common trace proto definition.
- [Span](https://github.com/census-instrumentation/opencensus-proto/blob/master/gen-go/trace/v1/trace.pb.go#L135) adds Node as an optional field.

This would allow for collector or agent services to batch the sending of spans from an arbitrary number of services.

### Impact on current agent and client libraries:
Since this field is optional, it would not prevent current optimizations (such as the pre-sending of Node data) from working as-is.

SergeyKanzhelev:
Should there be a containing object like `Context` that can be extended in future with additional metadata about individual spans? As we are talking about logs with strongly typed metadata - every log message will need to carry metadata like [this](https://github.com/elastic/ecs). And we may want to use some of it for spans as well.

CC: @mtwo

bogdandrutu:
From what I am seeing here Node is miss-used here. Node was added as an internal message to be used by library->agent->collector to know who generated the Span (I know that some data in the Node can be determined from the RPC connection, but that is not possible for example in the collector because the library talks to the agent not to the collector directly).

What you probably need is the "Resource" which main scope is to "logically" (pod, container, task number, service name, etc.) identify the task in the system. I think having the Resource in the Span is fine.

Maybe I don't understand what you try to achieve here but at least this is my mental model.

@yurishkuro  What do you have in "Process" entry?

maynk:
[...]

PR #174, opened 2019-01-23
Add Resource to Span

sjkaris:
As discussed in the wider meeting, adds Resource information to individual spans.

### DIFF
--- a/src/opencensus/proto/trace/v1/trace.proto
+++ b/src/opencensus/proto/trace/v1/trace.proto
@@ -300,6 +300,11 @@ message Span {
   // An optional number of child spans that were generated while this span
   // was active. If set, allows an implementation to detect missing child spans.
   google.protobuf.UInt32Value child_span_count = 13;
+
+  // An optional resource that this span describes. If not set, this span 
+  // should be part of a batch that does include the span information, unless resource 
+  // information is unknown.
+  opencensus.proto.resource.v1.Resource resource = 14;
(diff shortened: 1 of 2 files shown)


sjkaris:
fixes #169

src/opencensus/proto/trace/v1/trace.proto
```diff
@@ -290,6 +291,11 @@ message Span {
   // Status.Ok (code = 0).
   Status status = 11;
 
+  // An optional resource that this span describes. If not set, this span 
+  // should be part of a batch that does include the span information, unless resource
```
yurishkuro: "that does include the span information" or "that does include the **resource** information"?

sjkaris: ah you are right, should be resource. -- fixing

src/opencensus/proto/trace/v1/trace.proto
```diff
@@ -290,6 +291,11 @@ message Span {
   // Status.Ok (code = 0).
   Status status = 11;
 
+  // An optional resource that this span describes. If not set, this span
```
rghetia: span doesn't describe resource. May be this should be rephrased to something like "An optional resource to which this span is associated with".

sjkaris: Reworded. Please take another look

PR #181, opened 2019-02-27
Document that all maximum values must be specified.

### LINKED ISSUE #179: Can't tell the difference between unset and zero max_number_of_annotations.
https://github.com/census-instrumentation/opencensus-proto/blob/c9cc5de7251f8a320dd6cb4f92e5698ba86f284c/src/opencensus/proto/trace/v1/trace_config.proto#L41

This means that setting a `TraceConfig { constant_sampler { decision: true } }` sets all maximums to zero and produces not very interesting spans as a result.

g-easy:
Fixes #179.

### DIFF
--- a/src/opencensus/proto/trace/v1/trace_config.proto
+++ b/src/opencensus/proto/trace/v1/trace_config.proto
@@ -23,6 +23,7 @@ option java_outer_classname = "TraceConfigProto";
 option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1";
 
 // Global configuration of the trace service.
+// All maximum values must be specified, or the maximums will be set to zero.
 message TraceConfig {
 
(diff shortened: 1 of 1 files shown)


src/opencensus/proto/trace/v1/trace_config.proto
```diff
@@ -23,6 +23,7 @@ option java_outer_classname = "TraceConfigProto";
 option go_package = "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1";
 
 // Global configuration of the trace service.
+// All maximum values must be specified, or the maximums will be set to zero.
```
bogdandrutu: I would suggest to say:

"All fields must be specified, or the default zero values will be used for each type" - I think proto calls them zero values.

g-easy: How about this?

PR #182, opened 2019-02-27
Constant sampler: add option to always follow the parent's decision.

### LINKED ISSUE #180: Add a new constant sampler
Currently we support only two types of ConstantSampler: "ALWAYS_ON" and "ALWAYS_OFF" we should probably support a new one "ALWAYS_PARENT_DECISION" - which acts like "ALWAYS_OFF" if no parent.

g-easy:
Fixes #180.

### DIFF
--- a/src/opencensus/proto/trace/v1/trace_config.proto
+++ b/src/opencensus/proto/trace/v1/trace_config.proto
@@ -55,12 +55,22 @@ message ProbabilitySampler {
   double samplingProbability = 1;
 }
 
-// Sampler that makes a constant decision (either always "yes" or always "no")
-// on span sampling.
+// Sampler that always makes a constant decision on span sampling.
 message ConstantSampler {
 
-  // Whether spans should be always sampled, or never sampled.
+  // DEPRECATED: whether spans should be always sampled, or never sampled.
   bool decision = 1;
+
+  // How spans should be sampled:
+  // - Always on
+  // - Always off
+  // - Always follow the parent Span's decision (off if no parent).
+  enum ConstantDecision {
+    ALWAYS_OFF = 0;
+    ALWAYS_ON = 1;
+    ALWAYS_PARENT = 2;
+  }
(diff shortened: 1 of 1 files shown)


src/opencensus/proto/trace/v1/trace_config.proto
```diff
@@ -55,12 +55,22 @@ message ProbabilitySampler {
...
-  // Whether spans should be always sampled, or never sampled.
+  // DEPRECATED: whether spans should be always sampled, or never sampled.
   bool decision = 1;
+
+  // How spans should be sampled:
+  // - Always on
+  // - Always off
+  // - Always follow the parent Span's decision (off if no parent).
+  enum ConstantDecision {
+    ALWAYS_OFF = 0;
```
g-easy: What do you think about adding UNSET = 0;

This is for backwards compatibility: if decision_v2 is "unset" (i.e. zero default due to proto3), then decision_v1 takes precedence.

src/opencensus/proto/trace/v1/trace_config.proto
```diff
@@ -55,12 +55,15 @@ message ProbabilitySampler {
...
-// on span sampling.
+// Sampler that always makes a constant decision on span sampling.
 message ConstantSampler {
 
   // Whether spans should be always sampled, or never sampled.
   bool decision = 1;
+
+  // Whether to follow the parent's decision. If there is no parent, use
+  // the decision specified above.
+  bool follow_parent = 2;
```
g-easy: How about this option: just add another bool.

bogdandrutu: I think we don't use this yet, what about we just break compatibility and do the enum version?

g-easy: I'm okay with that.

g-easy:
PTAL.

PR #167, opened 2019-01-08
Start 0.3.0 development cycle

songy23:
Updates https://github.com/census-instrumentation/opencensus-proto/issues/161.

PR #183, opened 2019-03-07
Revert "Start 0.3.0 development cycle"

songy23:
Reverts census-instrumentation/opencensus-proto#167.

We haven't made the v0.2.0 release yet. Revert this change to start over the release process.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ apply plugin: 'maven'
 apply plugin: "signing"
 
 group = "io.opencensus"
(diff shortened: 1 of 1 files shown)


bogdandrutu:
What do you want to do with this PR?

songy23:
I still prefer to revert this commit and create a new one, since the old commit (https://github.com/census-instrumentation/opencensus-proto/commit/853d694fb7dbcedb5f2c884fc5a5ac1fb449445f) is on Jan 15. On that date we didn't make a release so it shouldn't represent the beginning of v0.3.0 cycle.

PR #187, opened 2019-03-13
Update gen-go files.

songy23:


songy23:
Apologies - need to run this first.

icco:
@songy23 not sure why, but this broke my builds:

```
../../../../pkg/mod/contrib.go.opencensus.io/exporter/stackdriver@v0.9.1/metrics.go:357:17: metric.GetName undefined (type *"github.com/census-instrumentation/opencensus-proto/gen-go/metrics/v1".Metric has no field or method GetName)
```

songy23:
@icco That's caused by https://github.com/census-instrumentation/opencensus-proto/pull/184 which removed the `name` field from metric, while opencensus-go-exporter-stackdriver used to depend on an unreleased version with that field. Filed https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/issues/101 and I'll take a look at it.

PR #195, opened 2019-03-26
Re-generate gen-go files with protoc 3.7.0.

songy23:
Re-generate gen-go files with the latest release of protoc. I suspect this is the cause of https://github.com/census-instrumentation/opencensus-service/issues/502.


songy23:
Not a priority now, since the gRPC issue has been fixed. Gen files will be updated in the next release.

PR #203, opened 2019-07-12
Preview: Use gogoproto to generate Go code

owais:
This generates the Go code with gogoproto instead of the default Go protobuf implementation. When merged, this should be released as the next major version as it can break the Go programs depending on this library.


tigrannajaryan:
@owais the build failed, is this expected?

owais:
The build seems to be failing because of a newer version of bazel that has deprecated some features. I'll try to look into it tomorrow.

songy23:
@owais https://github.com/census-instrumentation/opencensus-proto/pull/206 should fix it.

owais:
Fixed the travis build in: https://github.com/census-instrumentation/opencensus-proto/pull/207

PR #206, opened 2019-07-17
Remove grpc_java_repositories from Bazel build.

songy23:
For Java we use Gradle to generate and upload the jar, and other libraries should depend on https://search.maven.org/artifact/io.opencensus/opencensus-proto/. Bazel build rules is not very useful. In addition `grpc_java_repositories` has caused build failures in Travis (e.g https://travis-ci.org/census-instrumentation/opencensus-proto/jobs/557790386).


songy23:
Looks like there're other errors - closing for now.

PR #207, opened 2019-07-17
Fix travis build

owais:
Fixes travis build by pinning bazel to a known working version.

PR #210, opened 2019-07-23
Bring back bazel cache to make CI faster.

g-easy:
This was added in #176 then reverted in #207.


g-easy:
Before: [7m50s](https://travis-ci.org/census-instrumentation/opencensus-proto/builds/562417404?utm_source=github_status&utm_medium=notification)
After: [1m33s](https://travis-ci.org/census-instrumentation/opencensus-proto/builds/562454033?utm_source=github_status&utm_medium=notification)

PR #212, opened 2019-07-23
Add C++ targets, especially for gRPC services.

g-easy:


g-easy:
This PR needs #211 to be merged first, because we need to import the non-Java grpc repository to get `cc_grpc_library()`

g-easy:
cc @panzhongxian

PR #214, opened 2019-08-26
Remove grpc_java dependency and java_proto rules.

g-easy:
This is so that Envoy can consume opencensus-proto without adding
a dependency on grpc_java.


g-easy:
Motivating change in envoy is: https://github.com/envoyproxy/envoy/pull/8023

PR #218, opened 2020-07-17
Generate Go types using  protocolbuffers/protobuf-go

nilebox:
Updating Collector to golang/protobuf v1.4.2 leads to performance degradation in OpenCensus receiver, see https://github.com/open-telemetry/opentelemetry-collector/pull/1308#issuecomment-659206647

Switching to the newer `protoc-gen-go` seems to fix the issue.
Note that this is a completely different `protoc-gen-go` now:
- Currently https://github.com/golang/protobuf/tree/master/protoc-gen-go is being used
- This PR switches to https://github.com/protocolbuffers/protobuf-go/tree/master/cmd/protoc-gen-go

/cc @rghetia @bogdandrutu @james-bebbington


gen-go/agent/common/v1/common.pb.go
```diff
@@ -1,25 +1,47 @@
...
 // source: opencensus/proto/agent/common/v1/common.proto
 
+// NOTE: This proto is experimental and is subject to change at this point.
+// Please do not use it at the moment.
+
 package v1
 
 import (
-	fmt "fmt"
 	proto "github.com/golang/protobuf/proto"
```
nilebox, line 27: Apparently new generator still uses `"github.com/golang/protobuf/proto"` just not for marshaling.

src/install-protoc.sh
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env bash
+
```
nilebox, line 2: Added installation script based on https://github.com/open-telemetry/opentelemetry-proto/blob/ba6a2594e92252a0a75f6496d32d1b1c60f858c6/install-proto.sh

src/install-protoc.sh
```diff
@@ -0,0 +1,146 @@
...
+set -e
+
+readonly PROTOC_VER="3.12.3"
+readonly PROTOC_ARCHIVE="protoc-${PROTOC_VER}-linux-$( uname -m ).zip"
+readonly PROTOC_URL="https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VER}/${PROTOC_ARCHIVE}"
+
+readonly PROTOC_GEN_GO_VER="1.25.0"
+readonly PROTOC_GEN_GO_ARCHIVE="protoc-gen-go.v${PROTOC_GEN_GO_VER}.linux.amd64.tar.gz"
+readonly PROTOC_GEN_GO_URL="https://github.com/protocolbuffers/protobuf-go/releases/download/v${PROTOC_GEN_GO_VER}/${PROTOC_GEN_GO_ARCHIVE}"
```
nilebox, line 11: This is the main change in this PR - previously https://github.com/golang/protobuf/tree/master/protoc-gen-go was used

nilebox:
Please don't merge, I will try an alternative approach with generating Go types in the collector repo instead.
This will allow us to manage collector code independently from OpenCensus.

nilebox:
> I will try an alternative approach with generating Go types in the collector repo instead

It requires too much work that's probably not worth it, since `opencensus-proto` types are used by various components:
- The "old" interfaces use OpenCensus protos: https://github.com/open-telemetry/opentelemetry-collector/blob/master/consumer/consumerdata/consumerdata.go
- OpenCensus exporter is using the external OpenCensus exporter internally, which relies on `opencensus-proto` types: https://github.com/open-telemetry/opentelemetry-collector/blob/4046234701ae1a17cbfa428a86cf5b239f68cd74/exporter/opencensusexporter/opencensus.go#L35
- Self-observability package in the collector is using OpenCensus pipeline: https://github.com/open-telemetry/opentelemetry-collector/blob/master/obsreport/observability.go

Given that most of that code is to be eventually deleted from the repo or migrated to the "new" interfaces, it doesn't seem worth doing all that work.

PR #223, opened 2021-05-04
update to grpc-gateway v2

drewwells:
gRPC Gateway went through a major rewrite to v2 last year. The main impact that I see on this repo is the protobuf annotation changes from swagger to openapiv2 and of course the protoc plugin was renamed as well. I'll take care of those in this PR.

Ref: https://grpc-ecosystem.github.io/grpc-gateway/docs/development/grpc-gateway_v2_migration_guide/

### DIFF
--- a/src/install-protoc.sh
+++ b/src/install-protoc.sh
@@ -13,7 +13,7 @@ readonly PROTOC_GEN_GO_URL="https://github.com/protocolbuffers/protobuf-go/relea
 readonly PROTOC_GEN_GO_GRPC_VER="v1.30.0"
 readonly PROTOC_GEN_GO_GRPC_REPO_URL="https://github.com/grpc/grpc-go"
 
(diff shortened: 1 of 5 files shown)


jsuereth:
Thanks for looking at this!

src/install-protoc.sh
```diff
@@ -28,12 +28,13 @@ get() {
     local url="$1"
     local dest="$2"
 
-    if [[ -w $dest ]]
+    if [[ ! -w $(dirname $dest) ]]
```
drewwells, line 31: I had issues with this line calculating sudo correctly. I believe `-w` only evaluates valid paths. Since the destination does not exist, this always returned false. The result was `sudo wget` laid down a file as root:root and chmod below ran without sudo and was unable to change the file bits.

bogdandrutu:
@jsuereth you can close this, already done. See also https://github.com/census-instrumentation/opencensus-proto/pull/228

